PR #4, opened 2021-08-04
Multi-line dict improvements

jonyoder:
Tweaks a bunch of dict declarations to use multiple lines.


jonyoder:
Closing since older versions of helm don't support multi-line things.

PR #7, opened 2021-08-04
Rstudio pm public readme

kfeinauer:
Attempts to give similar treatment to RSPM as was done for RSW in regards to the README.

I am not very familiar with running/administering RSPM, so I probably missed something. Let me know if I should add anything for required settings, and let me know if I should add any recommended settings as well.

Associated with https://github.com/rstudio/sol-eng-helm/issues/64


charts/rstudio-pm/README.md.gotmpl
```diff
@@ -7,6 +7,18 @@
...
+## Required Configuration
+
+This chart requires the following in order to function:
+
+* A license key, license file, or address of a running license server. See the `license` configuration below.
+* A Kubernetes [PersistentVolume](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) that contains the data directory for RSPM.
+  * If `sharedStorage.create` is set, a PVC that relies on the default storage class will be used. Most Kubernetes environments do not have a default
+    storage class that you can use with `ReadWriteMany` access mode out-of-the-box. In this case, we recommend you disable `sharedStorage.create` and 
+    create your own `PersistentVolume` and `PersistentVolumeClaim`, then mount them into the container by specifying the `pod.volumes` and `pod.volumeMounts` parameters.
+  * If you cannot use a `PersistentVolume` to properly mount your data directory, you'll need to manually mount your NFS server in the container during startup
```
jonyoder, line 19: @colearendt have we experimented with configuring RSPM to run with S3 for its shared data store from a Helm chart install?

jonyoder (approved):
Looks good to me. It would be good to have @colearendt take a look, too, in particular at the shared volume notes.

charts/rstudio-pm/README.md.gotmpl
```diff
@@ -7,6 +7,18 @@
 
 {{ template "rstudio.install" . }}
 
+## Required Configuration
+
+This chart requires the following in order to function:
+
+* A license key, license file, or address of a running license server. See the `license` configuration below.
+* A Kubernetes [PersistentVolume](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) that contains the data directory for RSPM.
```
colearendt, line 15: I think we may need to clean up our terminology here (although I admit my usage of these terms is not very clean).

The way that the chart usually works is to set up a PVC with an appropriate storage class, which _generates_ the PV "behind the scenes." 

https://stackoverflow.com/a/48956790/6570011

In fact, I believe sometimes a user will "have access to create a PVC," but "not create a PV," because the storage class does that.

charts/rstudio-pm/README.md.gotmpl
colearendt: "manually mount your NFS server to the container during startup" can be done with an NFS volume (i.e. using `pod.volumes` and `pod.volumeMounts` directly without a PV/PVC)

https://kubernetes.io/docs/concepts/storage/volumes/#volume-types

@jonyoder Yes we have! In fact, this is what we do for https://colorado.rstudio.com/rspm/ - so that is definitely a case we should mention for RSPM

colearendt (changes requested):

kfeinauer:
@jonyoder @colearendt Updated to make the volume wording more accurate and also add a blurb about S3 storage for RSPM. Please have a look.

charts/rstudio-pm/README.md.gotmpl
```diff
@@ -19,6 +19,26 @@ This chart requires the following in order to function:
     mount them into the container by specifying the `pod.volumes` and `pod.volumeMounts` parameters.
   * If you cannot use a `PersistentVolume` to properly mount your data directory, you'll need to mount your data in the container
     by using a regular [Kubernetes Volume](https://kubernetes.io/docs/concepts/storage/volumes), specified in `pod.volumes` and `pod.volumeMounts`.
+  * Alternatively, S3 storage can be used. See the next section for details.
+
+## S3 Configuration
+
+Package Manager can be configured to store data in S3 buckets (see the [Adming Guide](https://docs.rstudio.com/rspm/admin/files-directories/#data-destinations)
```
jonyoder, line 26: s/Adming/Admin

charts/rstudio-pm/README.md.gotmpl
```diff
@@ -19,6 +19,26 @@ This chart requires the following in order to function:
...
+for more details). When configured this way, AWS access credentials are required. You must set the `awsAccessKeyId` and `awsSeretAccessKey` chart values
```
jonyoder, line 27: The property for the secret is misspelled, although I don't know if it's only misspelled here or globally.

kfeinauer: Thanks Jon. Fixed it up. My keyboard seems to drop keystrokes fairly frequently 😕  I need to get spellcheck working (can be a pain on Linux, but it's time).

jonyoder (approved):

colearendt (approved):
LGTM!

PR #9, opened 2021-08-04
Update index.yaml after rebuilding it

jonyoder:
Updates index.yaml for `helm.rstudio.com/index.yaml` after rebuilding it.

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -3,13 +3,13 @@ entries:
   rstudio-launcher-rbac:
   - apiVersion: v2
     appVersion: 0.2.1
-    created: "2021-08-04T16:39:16.142779-04:00"
+    created: "2021-08-05T08:04:52.345908-04:00"
     dependencies:
     - name: rstudio-library
       repository: file://../rstudio-library
       version: 0.1.13-rc01
     description: RBAC definition for the RStudio Job Launcher
-    digest: 0ab9c2be6e5ded80108cc95c89507cb0faee262d8429f39f4638b7906b78ca14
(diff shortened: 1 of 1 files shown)


colearendt:
Do we have two entries for rstudio-pm-0.2.0-rc03? Any ideas for why that is?

jonyoder:
@colearendt I noticed that and I have no idea why!

jonyoder:
Oh, I'll bet it appended it since the existing `index.yaml` already had it.

jonyoder:
@colearendt The problem was that we were using `https://helm.rstudio.com` as the helm repo when generating the index. This grabbed the existing index.yaml and appended all the rebuilt packages to it. I changed the default to `https://rstudio.com` instead so the index gets rebuilt from nothing. It's annoying since it must be a valid domain but must 404.

colearendt (approved):
Very strange!! Sounds like chart-releaser has some work to do to make it more idempotent 😬 

In any case, thanks for doing this digging! Now we know a lot more about how chart-releaser works, which is encouraging since we are building on top of it!

jonyoder:
I realized that the digest values are changing, and won't match the SHA256 checksums of the files on Github. I'll work on an alternate approach for the script.

colearendt:
Maybe this is the reason that cr is not idempotent? Maybe its digests are not consistent either? I don’t know enough about how that index.yaml is generated, I don’t think.

jonyoder:
It should be good now!

colearendt:
Well done! LGTM!

PR #20, opened 2021-08-05
Connect public readme

kfeinauer:
For now just added the same blurb about shared storage and licensing.

Is there anything else of note in the Connect configuration? I didn't see anything worth talking about.


jonyoder (approved):

colearendt (approved):
LGTM!

kfeinauer:
The status checks don't seem to be appearing correctly, even though I can see they are passed in Github Actions...

colearendt:
Yeah, I think this is a GitHub Actions bug b/c it only picks up the latest commit's jobs 😢  (and apparently the github actions user doesn't re-trigger other actions, probably to avoid circular control loops)

Issue #16, opened 2021-08-05
Allow "sharedStorage" mounting for RSPM / RSW with existing PVC


colearendt:
RSC allows this pattern. We should ensure the other charts are consistent

PR #22, opened 2021-08-06
Shared Storage Mounting for RSPM

jonyoder:
Connected to #16. Note that #16 covers shared storage mounting for both RSPM and RSW. This PR covers only RSPM, so another PR must follow for RSW eventually.


kfeinauer (approved):
Approved with the small caveat mentioned in your other PR.

kfeinauer (approved):
I think that sounds good!

PR #25, opened 2021-08-06
Update READMEs to clarify that the Helm config value gets mapped to the specific service configuration file

kfeinauer:
I think what was previously written could confuse users. Hopefully this wording is more clear.

It was a bit more obvious for the workbench chart because it specifically calls out which configuration files are needed and which config values generate them.

Associated with https://github.com/rstudio/sol-eng-helm/issues/64


charts/rstudio-connect/README.md.gotmpl
```diff
@@ -36,7 +36,7 @@ so you can set the database password with something like:
 ... --set config.Postgres.Password=mypassword ...
 ```
 
-The values are converted into configuration files in the necessary format via go-templating.
+TThe Helm `config` values are converted into the `rstudio-connect.gcfg` service configuration file via go-templating.
```
jonyoder, line 39: double `T` in `TThe`.

jonyoder (approved):
Looks great after one spelling nit!

colearendt (approved):
Nice!

PR #28, opened 2021-08-10
Fix the workbench test by disabling various settings that should be disabled for the test pod.

kfeinauer:
The previous method of calling set and passing . separated names for subkeys does not work. Tested `main` after https://github.com/rstudio/helm/pull/27 was merged and saw that the test pod was still hanging because it was not properly disabling the readiness probe and the exporter container.

This method creates the proper objects for all disabled objects and combines them into the `overrideDictionary.Values` to ensure everything is overridden properly. `mergeOverwrite` is used to ensure that we forcefully disable these settings as they should NEVER be set within a test pod.

### DIFF
--- a/charts/rstudio-workbench/templates/tests/test-verify-installation.yaml
+++ b/charts/rstudio-workbench/templates/tests/test-verify-installation.yaml
@@ -16,8 +16,11 @@ spec:
   {{- $overrideDict := . | deepCopy }}
   {{- $_ := set $overrideDict.Values "enableDiagnostics" true }}
   {{- $_ := set $overrideDict.Values "userCreate" true }}
-  {{- $_ := set $overrideDict.Values "readinessProbe.enabled" false }}
-  {{- $_ := set $overrideDict.Values "livenessProbe.enabled" false }}
-  {{- $_ := set $overrideDict.Values "startupProbe.enabled" false }}
-  {{- $_ := set $overrideDict.Values ".Values.prometheusExporter.enabled" false }}
+  {{- $disabledObject := dict "enabled" false }}
+  {{- $readinessProbe := dict "readinessProbe" $disabledObject }}
+  {{- $livenessProbe := dict "livenessProbe" $disabledObject }}
+  {{- $startupProbe := dict "startupProbe" $disabledObject }}
+  {{- $prometheusExporter := dict "prometheusExporter" $disabledObject }}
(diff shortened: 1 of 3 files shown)


colearendt (commented):
What if we refactored this whole thing into a big `mergeOverwrite`? That is the pattern we use elsewhere (with different precedence, because it is more user-customizable).

I think using `mergeOverwrite` with `set` is a bit confusing, and it's not clear to me how `set` behaves when there are conflicts / children / etc. I think building an object and then `mergeOverwriting` with `override.Values` may be more clear?

kfeinauer:
@colearendt Done.

charts/rstudio-workbench/templates/tests/test-verify-installation.yaml
```diff
@@ -13,11 +13,12 @@ spec:
...
+  {{- $startupProbe := dict "startupProbe" $disabledObject }}
+  {{- $prometheusExporter := dict "prometheusExporter" $disabledObject }}
   {{- $overrideDict := . | deepCopy }}
-  {{- $_ := set $overrideDict.Values "enableDiagnostics" true }}
-  {{- $_ := set $overrideDict.Values "userCreate" true }}
-  {{- $_ := set $overrideDict.Values "readinessProbe.enabled" false }}
-  {{- $_ := set $overrideDict.Values "livenessProbe.enabled" false }}
-  {{- $_ := set $overrideDict.Values "startupProbe.enabled" false }}
-  {{- $_ := set $overrideDict.Values ".Values.prometheusExporter.enabled" false }}
+  {{- $_ := mergeOverwrite $overrideDict.Values $prometheusExporter $startupProbe $livenessProbe $readinessProbe $topLevelParams }}
```
colearendt, line 23: This looks awesome! Does this `$_` syntax work with `mergeOverwrite`? Should we Create a `Values {rest of things}` dict and set this output to `$overrideDict`?

kfeinauer: This works as is. `mergeOverwrite` actually overrides the sub values and thus pollutes the `overrideDict.Values`. 

Also, it is invalid syntax to do what you suggested (already tried something similar - Go templating won't let you do that).

colearendt: Ah cool! Good to know, thanks!! 🏆 TIL! 😄

colearendt (approved):
LGTM!

PR #29, opened 2021-08-12
Tweak profiles

colearendt:
- improve whitespace consistency
- handle a bunch of rough edges
  - duplicate session config to the target namespace (if different from release namespace)
  - change debug logging to off by default
  - ensure launcher / target namespace is passed properly to the launcher (when different)
  - fix profiles / overrides handling of defaults (i.e. mount session configmap by default)
  - add rbac that is needed for the sidecar container (when launching sessions into a different namespace)
  - change default value for session-callback-address (and set programmatically)
  - remove the launcher service - it does not work with a single set of pods because readiness probes cannot support multiple services (probably by design). So we will remove it for now

### DIFF
--- a/charts/rstudio-workbench/templates/configmap-general.yaml
+++ b/charts/rstudio-workbench/templates/configmap-general.yaml
@@ -16,9 +16,7 @@
 {{- /* Define the default rserver values - point to the launcher and IDE services in Kubernetes */}}
 {{- $defaultIDEServiceName := include "rstudio-workbench.fullname" . }}
 {{- $defaultIDEServiceURL := printf "http://%s.%s.svc.cluster.local:80" $defaultIDEServiceName $.Release.Namespace }}
-{{- $defaultLauncherServiceName :=  print "%s-launcher" (include "rstudio-workbench.fullname" .) }}
-{{- $defaultLauncherServiceURL := printf "%s.%s.svc.cluster.local" $defaultLauncherServiceName $.Release.Namespace }}
-{{- $defaultRServerConfigValues := dict "launcher-address" ($defaultLauncherServiceURL) "launcher-port" ("80") "launcher-sessions-callback-address" ($defaultIDEServiceURL) }}
+{{- $defaultRServerConfigValues := dict "launcher-sessions-callback-address" ($defaultIDEServiceURL) }}
 {{- $defaultRServerConfig := dict "rserver.conf" ($defaultRServerConfigValues) }}
 ---
 apiVersion: v1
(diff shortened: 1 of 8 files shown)


charts/rstudio-workbench/templates/configmap-general.yaml
```diff
@@ -12,11 +12,12 @@
...
 {{- $defaultProfiles := dict "default-container-image" (first $defaultImages) "container-images" $defaultImages "allow-unknown-images" 1 }}
-{{- if $defaultOverrides }}
-  {{/* if defaultOverrides exist - add them */}}
-  {{- $defaultProfiles = mergeOverwrite $defaultProfiles ( dict "job-json-overrides" $defaultOverrides ) }}
-{{- end }}
 {{- $defaultProfilesConfig := dict "*" $defaultProfiles }}
+{{- /* Define the default rserver values - point to the launcher and IDE services in Kubernetes */}}
+{{- $defaultIDEServiceName := include "rstudio-workbench.fullname" . }}
+{{- $defaultIDEServiceURL := printf "http://%s.%s.svc.cluster.local:80" $defaultIDEServiceName $.Release.Namespace }}
+{{- $defaultRServerConfigValues := dict "launcher-sessions-callback-address" ($defaultIDEServiceURL) }}
```
kfeinauer, line 19: This solves an issue that @nahara7 was looking into to fix the callback address in general so that the verify-installation test works correctly!

charts/rstudio-workbench/templates/configmap-session.yaml
```diff
@@ -5,5 +5,17 @@ metadata:
   name: {{ include "rstudio-workbench.fullname" . }}-session
   namespace: {{ $.Release.Namespace }}
 data:
-{{ include "rstudio-library.config.ini" .Values.config.session | indent 2 }}
+  {{- include "rstudio-library.config.ini" .Values.config.session | nindent 2 }}
+
+{{- /* if session/target namespace differs - duplicate session configuration there */}}
```
kfeinauer, line 10: What values do we need? What scenario are we covering? Could we add this to the comments here?

colearendt:
@jonyoder RSW _can_ work without the launcher, but I am talking about the "Kubernetes Service." It's not exactly usable (per message on slack), so I am of the opinion that we should remove it and just route to the launcher over loopback

charts/rstudio-workbench/templates/configmap-session.yaml
colearendt: Yes! I'll add color to the comment. The "session" config is mounted onto the session pods automatically per https://github.com/rstudio/helm/blob/27cb5989fde1c6b1049e515f0d2b87b8560d01a9/charts/rstudio-workbench/values.yaml#L10 and https://github.com/rstudio/helm/blob/27cb5989fde1c6b1049e515f0d2b87b8560d01a9/charts/rstudio-workbench/templates/configmap-general.yaml#L6-L12

This breaks if "targetNamespace" != "release namespace" because configmaps cannot be mounted across namespaces

PR #48, opened 2021-09-10
Fix rsw

colearendt:
Add news
Fix a pernicious little issue that messed up our load-balancer file!

(We were using a different key than the "selectorLabel" if your `.Release.Name` was not `rstudio-workbench`)

https://github.com/rstudio/helm/blob/bf49d985b6056eb8f94e6fdd81ba86969fe283be/charts/rstudio-workbench/templates/_helpers.tpl#L271

(Versus "fullname" below)


colearendt:
Verified that it resolves the issue!

PR #50, opened 2021-09-15
remove/update python configurations in example connect values

dbkegley:
depends on #49 

- enables python by default when `launcher.enabled=true`
- fixes config precedence when merging `.Values.config`
- removes python configurations from existing examples

Previously: 
```
$ helm template connect-test ./ --set config.Launcher.Kubernetes=false --set launcher.enabled=true --show-only templates/configmap.yaml | grep -A3 Launcher
    [Launcher]
    ClusterDefinition = /etc/rstudio-connect/runtime.yaml
    Enabled = true
    Kubernetes = true
```

Now:
```
$ helm template connect-test ./ --set config.Launcher.Kubernetes=false --set launcher.enabled=true --show-only templates/configmap.yaml | grep -A3 Launcher
    [Launcher]
    ClusterDefinition = /etc/rstudio-connect/runtime.yaml
    Enabled = true
    Kubernetes = false
```


colearendt (commented):
LGTM! Do you mind merging in the latest from main, adding a NEWS entry and bumping the version?

colearendt (approved):
Nvm - didn't realize this is just in examples!

I guess we have to wait til we make launcher the default before we can change the default values in the chart 😢 

One possibility - we could always inject "Python.Enabled = true" when launcher is enabled 🤷‍♂️ If it isn't set explicitly otherwise

dbkegley:
@colearendt I'm open to enabling python when launcher is also enabled, want me to handle that in this PR or open a new issue?

colearendt (approved):
LGTM! Merge at will!

PR #56, opened 2021-10-13
small fixes for svc.yaml and values.yaml

Aarhan:
The two main goals of this commit are to simplify the port configuration by adding the missing service port value to the values.yaml with a default value of 80, and also removing the hardcoded port 80 in the service template and replacing it with the $svcPort variable thats used in the ingress allowing the service port to be changed more easily.


Aarhan:
Hey!

I had actually done it this way as I saw it used in the ingress.yaml, but I have since learned this was due to the range operations further down in the yaml before the service port section (TIL). 

I added another commit that simply sets this to:

{{ .Values.service.port }}

as recommended!

Aarhan:
I think it was actually in my update to the NEWS.md for rstudio-pm. Tried a new commit that I think removes the trailing space.

PR #96, opened 2021-12-01
Fix health check

colearendt:
A customer environment had issues with `http_proxy` and `https_proxy` being set causing troubles with our health check. It is possible to set `no_proxy` environment variables (although `curl` does not seem to respect CIDR notation - [some discussion on ambiguity here](https://github.com/curl/curl/issues/1208)), but this will enable turning off the health check entirely or adding arguments to curl - i.e. `--noproxy`, etc.

[Worth noting that `healthz` is deprecated in favor of `livez` and `readyz`](https://kubernetes.io/docs/reference/using-api/health-checks/)

Also document and formalize the previously hidden `enableDiagnostics` flag.

Worth noting that all of this entrypoint script will likely go away once we switch to the new launcher with its ability to catch these values from the environment 🙈

### DIFF
--- a/charts/rstudio-workbench/templates/_helpers.tpl
+++ b/charts/rstudio-workbench/templates/_helpers.tpl
@@ -31,13 +31,21 @@ containers:
   {{- $defaultVersion := .Values.versionOverride | default $.Chart.AppVersion }}
   image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default $defaultVersion }}"
   env:
-  {{- if .Values.enableDiagnostics }}
+  {{- if .Values.diagnostics.enabled }}
   - name: DIAGNOSTIC_DIR
-    value: "/var/log/rstudio"
+    value: "{{ .Values.diagnostics.directory }}"
   - name: DIAGNOSTIC_ONLY
     value: "true"
   - name: DIAGNOSTIC_ENABLE
     value: "true"
+  {{- end }}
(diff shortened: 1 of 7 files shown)


charts/rstudio-workbench/templates/_helpers.tpl
```diff
@@ -31,14 +31,22 @@ containers:
...
   - name: DIAGNOSTIC_ONLY
     value: "true"
   - name: DIAGNOSTIC_ENABLE
     value: "true"
   {{- end }}
+  {{- if not .Values.launcher.kubernetesHealthCheck.enabled }}
+  - name: RSTUDIO_LAUNCHER_STARTUP_HEALTH_CHECK
+    value: disabled
+  {{- end }}
+  {{- if .Values.launcher.kubernetesHealthCheck.enabled }}
```
jonyoder, line 46: Why not simply use `{{ - else }}` here?

jonyoder (approved):
Looks like a great improvement to me!

charts/rstudio-workbench/templates/_helpers.tpl
colearendt: I think it was mostly in a desire to be declarative 🙈  They are not "technically" bound as if/else (i.e. right now our default on the HEALTH_CHECK header is to leave it undefined if things are enabled) so I thought it might be confusing to define them as such.

PR #98, opened 2021-12-01
Check storage class

### LINKED ISSUE #90: Is there a way to catch storage class access modes at deployment time?
Otherwise we leave the user a long debugging trail figuring out that we require `ReadWriteMany` by default

It'd be nice if there was a reliable way to see storage classes available for users and optionally fail

colearendt:
Close #90 

The motivation: if a user deploys our chart directly onto an EKS cluster setting `homeStorage.create=true`, they will get a not functioning server, a hard to debug situation (have to see that PVC provisioning failed and why), and a hard to fix situation (have to basically delete everything and start over once you have a storage class that supports ReadWriteMany or have set `accessModes` properly - both of which are immutable b/c PVCs cannot be changed once created).

It would be much nicer to give the user an error at install time when they go to use this storage class.

Unfortunately, Kubernetes does not "store" whether a given storage class supports a given access mode. This is documented [here](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes), but not able to be "looked up" in the Kbuernetes API

As a result, I do a name lookup for ones that we know cause trouble (i.e. `gp2`), as well as provisioners that we know cause trouble (i.e. `kubernetes.io/aws-ebs`).

It's possible that these storage classes change over time, and with different renamings / etc., it is impossible that we catch everything. However, I have seen enough users get tripped up by this to think that it would do us well to catch as many as we can.

Thoughts?

Examples:
```
# success
helm upgrade --install rsw-4 ./charts/rstudio-workbench --set homeStorage.storageClassName=gp2 --set homeStorage.create=true --set 'homeStorage.accessModes[0]'=ReadWriteOnce
Release "rsw-4" has been upgraded. Happy Helming!


# fail b/c specified gp2
helm upgrade --install rsw-4 ./charts/rstudio-workbench --set homeStorage.storageClassName=gp2 --set homeStorage.create=true
Error: UPGRADE FAILED: template: rstudio-workbench/templates/NOTES.txt:44:3: executing "rstudio-workbench/templates/NOTES.txt" at <include "rstudio-library.check-storage-class" (dict "accessModes" .Values.homeStorage.accessModes "storageClassName" .Values.homeStorage.storageClassName "key" "homeStorage" "check" true)>: error calling include: template: rstudio-workbench/templates/_helpers.tpl:439:12: executing "rstudio-library.check-storage-class" at <fail (printf "\n\nThe storage class '%s'%s is known to not support 'ReadWriteMany' PersistentVolumeClaims. This is important for proper disk provisioning. \nFix by setting `%s.accessModes` appropriately, changing `%s.storageClassName` or disabling this check with `%s.checkAccessModes: false`\n\n" $storageClassName $isDefault $key $key $key)>: error calling fail: HELM_ERR_START

The storage class 'gp2' is known to not support 'ReadWriteMany' PersistentVolumeClaims. This is important for proper disk provisioning. 
Fix by setting `homeStorage.accessModes` appropriately, changing `homeStorage.storageClassName` or disabling this check with `homeStorage.checkAccessModes: false`

HELM_ERR_END

# fail b/c default is gp2
✗ helm upgrade --install rsw-4 ./charts/rstudio-workbench  --set homeStorage.create=true
Error: UPGRADE FAILED: template: rstudio-workbench/templates/NOTES.txt:42:3: executing "rstudio-workbench/templates/NOTES.txt" at <include "rstudio-library.check-storage-class" .>: error calling include: template: rstudio-workbench/templates/_helpers.tpl:425:10: executing "rstudio-library.check-storage-class" at <fail (printf "\n\nThe storage class '%s'%s is known to not support ReadWriteMany PersistentVolumeClaims. \nFix by setting `accessModes` appropriately, changing `storageClassName` or disabling this check with `checkAccessModes: false\n\n`" $storageClassName $isDefault)>: error calling fail: HELM_ERR_START

The storage class 'gp2' (the default on this cluster) is known to not support 'ReadWriteMany' PersistentVolumeClaims. 
Fix by setting `homeStorage.accessModes` appropriately, changing `homeStorage.storageClassName` or disabling this check with `homeStorage.checkAccessModes: false`

HELM_ERR_END


```

**EDIT:** Plan is to refactor into the `rstudio-library` chart, although only rolling out for rstudio-workbench for now, so that we can test a bit before integrating with Connect and RSPM.

### DIFF
--- a/charts/rstudio-workbench/templates/NOTES.txt
+++ b/charts/rstudio-workbench/templates/NOTES.txt
@@ -38,3 +38,8 @@ NOTE: `.Values.config.server.rserver\.conf.server-license-type` is configured ma
 Please consider removing this configuration value.
 
 {{- end }}
+
+{{- /* check for invalid storage classes */ -}}
+{{- if .Values.homeStorage.create }}
+{{ include "rstudio-library.check-storage-class" (dict "accessModes" .Values.homeStorage.accessModes "storageClassName" .Values.homeStorage.storageClassName "key" "homeStorage" "check" true) }}
+{{- end }}
--- a/charts/rstudio-workbench/templates/_helpers.tpl
+++ b/charts/rstudio-workbench/templates/_helpers.tpl
@@ -396,3 +396,48 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- define "rstudio-workbench.xdg-config-dirs" -}}
 {{  trimSuffix ":" ( join ":" (list .Values.xdgConfigDirs (join ":" .Values.xdgConfigDirsExtra) ) ) }}
 {{- end -}}
+
+{{- /*
+  A helper to check whether the provided storage class is a problematic one...
+  Unfortunately, Kubernetes does not "store" what access modes are allowed for a storage class
+  So we just store a list of known troublemakers (and reference docs here: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes)
+
+  Argument Dict:
+    "accessModes": required. the access modes for the given storage pvc
+    "storageClassName": required. the storage class name in use (or `false` if using the default)
+    "key": optional. default "storage". The values key to use when referencing `accessModes` or `storageClassName`
+    "check": optional. default true. Whether to run the check or not
+*/ -}}
+{{- define "rstudio-library.check-storage-class" -}}
+  {{- $allowedAccessModes := .accessModes }}
+  {{- $storageClassName := .storageClassName }}
+  {{- $key := .key | default "storage" }}
+  {{- $check := .check | default true }}
+  {{- if $check }}
+    {{- $storageClass := dict }}
+    {{- $isDefault := "" }}
+    {{- $knownNotReadWriteManyProvisioners := list "kubernetes.io/aws-ebs" }}
+    {{- $knownNotReadWriteManyClasses := list "gp2" }}
+    {{- $allStorageClasses := lookup "storage.k8s.io/v1" "StorageClass" "" "" }}
+    {{- range $sc := $allStorageClasses.items }}
+      {{- if not $storageClassName }}
+        {{- /* need to get the default */ -}}
+        {{- if eq (get $sc.metadata.annotations "storageclass.kubernetes.io/is-default-class") "true" }}
+          {{- /* this is the default storage class */ -}}
+          {{- $storageClassName = $sc.metadata.name }}
+          {{- $storageClass = $sc }}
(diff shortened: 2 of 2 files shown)


atheriel:
It's hard for me to give a credible line-by-line review of the YAML changes, but I'll make the following comments:

* A static list of bad storage classes is not very elegant but it's fairly maintainable, in my view. Kubernetes providers don't introduce/change storage classes that often, and we could remove this workaround in favour of a more robust solution in the future without affecting users.

* Consider adding [`AzureDisk`](https://kubernetes.io/docs/concepts/storage/storage-classes/#azure-disk) here, it's the default storage class on Azure Kubernetes Service but does not support RWX -- a fact that bit me personally when I used to run AKS clusters.

As an aside: this is a really painful example of Helm's limitations. It's not really a programming language, so when you want do stuff like this it ends up being really opaque.

jonyoder (approved):
Whew! I can tell I've been away from Helm templates for a while. This looks good to me, although I really have very little knowledge in this arena.

atheriel:
As a follow up: I looked through the (deprecated, but still useful) `helm/stable` repo to see if I could find patterns for solving this. What I found is that many prominent charts -- wordpress, airflow, minio, sentry -- have exactly this requirement/issue but seem to solve it by having a large warning in the `README` telling users to make sure they use a RWX provisioner.

This makes me think there isn't an obvious, accepted solution in the Helm community for this problem, so that's even more of an argument for this hack.

colearendt:
Thanks for the awesome feedback, y'all! @atheriel do you know off-hand what the exact / default `metadata.name` of an `AzureDisk` is? Or what the `provisioner` is? I wanted to do `AzureDisk` too but wasn't sure what it was called and didn't want to spin up an AKS cluster to find out 🙈  (i.e. `AWSElasticBlockStore` is called `gp2` 😑 )

atheriel:
From the linked docs in my comment it *appears* the `provisioner` is `kubernetes.io/azure-disk`, and from my recollection the `metadata.name` is `azuredisk`. But I'm hazy on both. Microsoft has really good AKS docs, I'm sure we can find a tutorial with the details instead of spinning up a cluster.

d-m:
Could this check be done with a `pre-install, pre-upgrade` Helm hook that uses `kubectl` to attempt to create a test pvc? The `prometheus-operator` project gets creative with a [helm hook](https://github.com/helm/charts/blob/d58247103fafd5930792bb1fd3ac240547523200/stable/prometheus-operator/templates/prometheus-operator/cleanup-crds.yaml#L33) that uses `kubectl` to clean up CRDs.

PR #99, opened 2021-12-02
Add support for generating a ServiceMonitor CRD for RSPM

atheriel:
This is required in clusters that use the Prometheus Operator instead of a regular Prometheus server. It's also a very common feature of public Helm charts for applications with Prometheus metrics.

Note that there are many other fields [in this CRD][0] that I have not yet exposed via `values.yml`. In researching comparable Helm charts, I found that these fields were very inconsistently exposed, so I decided to start with the smallest useful subset instead.

Also, I personally never used any of the other fields when operating my own clusters, so I'm even more biased towards simplicity here.

~This is a draft PR because I wrote it all from memory and we should actually test it against a real cluster with a local Operator before merging.~

[0]: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#endpoint

### DIFF
--- a/charts/rstudio-pm/templates/servicemonitor.yaml
+++ b/charts/rstudio-pm/templates/servicemonitor.yaml
@@ -1,0 +1,21 @@
+{{- if and .Values.config.Metrics.Enabled .Values.serviceMonitor.enabled -}}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "rstudio-pm.fullname" . }}
+  namespace: {{ .Values.serviceMonitor.namespace | default .Release.Namespace }}
+  labels:
+    {{- include "rstudio-pm.labels" . | nindent 4 }}
+  {{- if .Values.serviceMonitor.additionalLabels }}
+    {{- toYaml .Values.serviceMonitor.additionalLabels | nindent 4 }}
+  {{- end }}
+spec:
+  endpoints:
+    - port: metrics
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace | quote }}
+  selector:
+    matchLabels:
+      {{- include "rstudio-pm.selectorLabels" . | nindent 6 }}
+{{- end }}
--- a/charts/rstudio-pm/values.yaml
+++ b/charts/rstudio-pm/values.yaml
@@ -147,6 +147,16 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+serviceMonitor:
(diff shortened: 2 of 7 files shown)


charts/rstudio-pm/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if and .Values.config.Metrics.Enabled .Values.serviceMonitor.enabled -}}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "rstudio-pm.fullname" . }}
+{{- if .Values.serviceMonitor.namespace }}
+  namespace: {{ .Values.serviceMonitor.namespace }}
```
colearendt, line 7: I personally am more fond of the pattern using `default` rather than `if/else` (worth noting that the ordering of `default` is always hard to get right 🙈 ):
```
namespace: {{ .Values.serviceMonitor.namespace | default $.Release.Namespace }}
```

charts/rstudio-pm/values.yaml
```diff
@@ -147,6 +147,16 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+serviceMonitor:
+  # -- If true, create a ServiceMonitor CRD for use with a Prometheus Operator
+  enabled: false
+  # -- Namespace to create the ServiceMonitor in (usually the same as the one in
+  # which the Operator is running). Defaults to the one used by the chart
+  namespace: null
```
colearendt, line 155: The pattern that we often use for this is `""`, because that keeps the type the same and allows `default` to still work

https://helm.sh/docs/chart_template_guide/function_list/#default

i.e. here:
https://github.com/rstudio/helm/blob/33f517f0c1b5015decfbf6a6424f182ff4e39ebe/charts/rstudio-connect/values.yaml#L179-L183

colearendt (commented):
Nice! One little nit, and a few thoughts:

- I oftentimes will just let CI fail by not bumping the version until we are ready to merge (to cut down on merge conflicts). Probably just a matter of personal preference though
- this is probably an anti-pattern, but depending on where the other "options"/values are, we could always just have a blob "escape hatch" with `toYaml` or some such:
https://github.com/rstudio/helm/blob/33f517f0c1b5015decfbf6a6424f182ff4e39ebe/charts/rstudio-connect/templates/deployment.yaml#L48

atheriel:
I don't mind rebasing to keep the merge noise down (as I've done for the latest round of fixes).

atheriel:
I've tested in cluster and it seems to be working :+1:

colearendt (approved):
LGTM! Very nice!

PR #103, opened 2021-12-02
Add support for adding arbitrary labels to RStudio Workbench pods

### LINKED ISSUE #101: rstudio-workbench pod metadata labels
Is there a way to add a metadata label to the pod template using the rstudio-workbench helm chart values.yaml?
This is what i need the workbench pod to look like:
```
apiVersion: v1
kind: Pod
metadata:
  name: demo
  labels:
    aadpodidbinding: <aad_pod_identity_name>
```

atheriel:
This should enable users that have additional integrations (like the Azure AD feature discussed in #101) that need to be injected via pod labels.

Closes #101.

### DIFF
--- a/charts/rstudio-workbench/values.yaml
+++ b/charts/rstudio-workbench/values.yaml
@@ -232,7 +232,7 @@ pod:
   volumes: []
   # -- volumeMounts is injected as-is into the "volumeMounts:" component of the pod.container spec
   volumeMounts: []
-  # -- podAnnotations is a map of keys / values that will be added as annotations to the rstudio-pm pods
+  # -- Additional annotations to add to the rstudio-workbench pods
   annotations: {}
   # -- Additional labels to add to the rstudio-workbench pods
   labels: {}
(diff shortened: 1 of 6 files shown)


charts/rstudio-workbench/values.yaml
```diff
@@ -234,6 +234,8 @@ pod:
   volumeMounts: []
   # -- podAnnotations is a map of keys / values that will be added as annotations to the rstudio-pm pods
```
colearendt, line 235:
```suggestion
  # -- pod.annotations is a map of keys / values that will be added as annotations to the rstudio-pm pods
```
holdover from a prior age 😅

colearendt (approved):
Looks awesome! 1 nit - maybe add an example to the `ci/complex-values.yaml` or somewhere, so that we are exercising this in CI / ensure it templates properly and satisfies linting?

colearendt:
Oh! And probably worth adding a `NEWS.md` entry as well!

atheriel:
@colearendt Should be all good now, provided the install test passes.

PR #111, opened 2021-12-10
rstudio-workbench startupUserProvisioning values option

### LINKED ISSUE #110: ability to configure supervisor sssd.conf
For user provisioning purposes, we require additional control on when sssd is started. Could it be possible to allow the edit of `rstudio-workbench-start-user`  configmap similar to the `rstudio-workbench-start-custom` configmap?
https://github.com/rstudio/helm/blob/1e67050ac2e328a1707193ef418b91ce142c4c34/charts/rstudio-workbench/templates/configmap-startup.yaml#L36-L54

What I'm looking for is a to change the supervisor sssd.conf, so that it does not launch when supervisor starts by setting
```
[program:sssd]
...
autostart=false
...
```
This way it can be launched in the correct order (since no program in supervisor can be restarted) by another supervisor program after executing e.g. a realm join
```
[program:realm-join]
command=/startup/custom/realm-join.sh && supervisorctl start sssd
autostart=true
numprocs=1
stdout_logfile=/dev/stdout
stdout_logfile_maxbytes=0
stdout_logfile_backups=0
stderr_logfile=/dev/stderr
stderr_logfile_maxbytes=0
stderr_logfile_backups=0
```

samssann:
Adds ability to configure supervisor sssd.conf #110 using the `config.startupUserProvisioning` values option

Close #110

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-workbench
 description: Official Helm chart for RStudio Workbench
-version: 0.5.0-rc12
+version: 0.5.0-rc13
 apiVersion: v2
 appVersion: 2021.09.0-351.pro6
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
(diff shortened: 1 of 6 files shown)


colearendt (commented):
This looks awesome! I like this approach better than the one I initially took 😅  Thanks for the PR!

Would you mind bumping the version to the next `rc`? I suspect that will get linting to pass - I can take care of the `helm-docs` dealio - it fails weirdly from forks 😞

samssann:
@colearendt thanks! I updated the version to rc-13.

PR #119, opened 2021-12-16
Use `k8s.networking.io/v1` Ingress API on Kubernetes 1.19+

### LINKED ISSUE #117: Add support for `networking.k8s.io/v1` `Ingress` resources
The line https://github.com/rstudio/helm/blob/02c1b295887a827aed71bf0a5118c40f2761b053/charts/rstudio-workbench/templates/ingress.yaml#L4 checks for Kubernetes version 1.14 or greater and uses `networking.k8s.io/v1beta1` for ingress if the check evaluates true.

A similar check should be added to use `networking.k8s.io/v1` if on Kubernetes 1.19 or later, since `networking.k8s.io/v1beta1` `Ingress` resources are deprecated in Kubernetes 1.19 and unavailable in Kubernetes 1.22.

Additional information:

* https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122
* https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#what-to-do

d-m:
This pull request provisions Ingress resources using `k8s.networking.io/v1` API when on Kubernetes 1.19+. This is particularly important for Kubernetes 1.22+, since prior Ingress APIs were removed in this version.

I added several helpers to `rstudio-library` to consolidate the logic and then updated `rstudio-pm`, `rstudio-connect` and `rstudio-workbench` charts to use the updated version. If this approach doesn't fit with the vision I'm happy to refactor.

Close #117


atheriel:
Thanks, this is a fantastic contribution.

d-m:
Linting was failing due to version bump:

```
 ✖︎ rstudio-connect => (version: "0.2.14", path: "charts/rstudio-connect") > Chart version not ok. Needs a version bump!
45
 ✖︎ rstudio-library => (version: "0.1.18", path: "charts/rstudio-library") > Chart version not ok. Needs a version bump!
46
 ✖︎ rstudio-pm => (version: "0.2.9", path: "charts/rstudio-pm") > Chart version not ok. Needs a version bump!
47
 ✖︎ rstudio-workbench => (version: "0.5.0-rc14", path: "charts/rstudio-workbench") > Chart version not ok. Needs a version bump!
```

Looking at recent tags, it looks like rstudio-workbench has release candidates and the others bump the semantic version directly. Since this is adding a new feature in a backward compatible way, I think that the following makes sense, but I'll wait for some guidance before I bump them:

* rstudio-connect 0.2.14 -> 0.3.0
* rstudio-library 0.1.18 -> 0.2.0
* rstudio-pm 0.2.9 -> 0.3.0
* rstudio-workbench 0.5.0 -> 0.6.0-rc1

colearendt:
Howdy @d-m ! Yes, this is an awesome contribution, thank you! - I am hoping to get to a more thorough review today, but you are right. For Workbench, we just released a bunch of breaking changes / etc. that we let bake for a while. I would suspect the best pattern forward is to sub-minor-version-bump each of the products, since IIUC this is backwards compatible, correct?:

- rstudio-connect 0.2.15 -> 0.2.16 (we just bumped to 0.2.15 b/c we bumped RSC's version)
- rstudio-library 0.1.18 -> 0.1.19
- rstudio-pm 0.2.9 -> 0.2.10
- rstudio-workbench 0.5.0 -> 0.5.1

Also, may be worth adding a `NEWS.md` entry for each product, although I realize that gets tedious 😅

colearendt (commented):
This is really good stuff! I still want to play with it a smidge before merging, but the only nitpick that I have is that several of the files are missing newline characters at the end - often editors will have a setting that configures whether to set that or not

image

d-m:
Sounds good. I'll bump the versions to the ones you specified, update the `NEWS.md` files, and add those newlines this afternoon Eastern time.

colearendt (approved):
Looks awesome! Thank you!!

(I hope you will excuse the randomness I added. CI fails weirdly when coming from forks, so I added a few helpers to help make running those CI tests locally easier, ran them, and fixed the silly issues they caused. I also added a few CI values to ensure that the ingress resources are running through `linting` properly 😄 )

PR #121, opened 2021-12-20
Simplify handling of Service annotations for RSPM and improve their documentation

atheriel:
I've added mention of [internal load balancer annotations](https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer) to the documentation, given that this is likely to be the most common reason to use this field.

In addition, the existing YAML helper is not actually necessary and has a confusing name, so I removed it.


atheriel:
Superceded by #122.

PR #122, opened 2021-12-20
Overhaul RSPM service generation

atheriel:
This PR contains the following changes, bundling the last one from #121:

* Use a `ClusterIP` service by default instead of a `NodePort`. See below.
* Add support for controlling [`loadBalancerIP`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) and [`clusterIP`](https://kubernetes.io/docs/concepts/services-networking/service/#choosing-your-own-ip-address), which are commonly-used service features.
* Ignore nodePort settings when the service type is not `NodePort`. Otherwise, this will generate server-side validation errors. It's debatable whether we should just let this issue bubble up to users, but the general conventions in the Helm community seem to lean towards "be liberal in what you accept", so that's what I'd go by here as well.
* Simplify handling of Service annotations. The existing YAML helper is not actually necessary and has a confusing name, so I removed it.
* Tweak documentation for `service.nodePort` and `service.annotations`.

Since the first of these is a breaking change, I bumped the minor version for the chart.

## Moving to ClusterIP

`ClusterIP` is the Kubernetes default, exposing the service in-cluster only. This is a secure, well-understood default. For external users, there are a few options:

* Use an `Ingress`. We support this already and it works with `ClusterIP`.

* Use a `LoadBalancer`. This can be expensive so it makes a poor default.

* Use a `NodePort` and some kind of external load balancing solution.

It seems like the last route (which is the current default) is much more uncommon than the first two, so I'd advocate for this change.

In addition:

* `NodePorts` are a very limited cluster resource, so we should avoid consuming them if we can avoid it.

* Defaulting to `NodePort` is a poor security choice, especially if we have users that don't understand the consequences.

### DIFF
--- a/charts/rstudio-pm/Chart.yaml
+++ b/charts/rstudio-pm/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-pm
 description: Official Helm chart for RStudio Package Manager
-version: 0.2.9
+version: 0.3.0
 apiVersion: v2
 appVersion: 2021.09.0-1
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
--- a/charts/rstudio-pm/NEWS.md
+++ b/charts/rstudio-pm/NEWS.md
@@ -1,3 +1,10 @@
+# 0.3.0
+
+- BREAKING: The generated service will now have type `ClusterIP` by default.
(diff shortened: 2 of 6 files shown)


colearendt (approved):
LGTM! This looks great. I am going to create an issue to follow up on this PR by tracing through some of these changes to other charts once they have had a while to "bake"

PR #128, opened 2022-01-03
Add extra deploy

### LINKED ISSUE #115: Install arbitrary resources with the Helm chart
My team has written a Helm chart internally to install RStudio Connect and Workbench and we would like to replace it with these. As part of our Helm charts, we install some other custom resources like cert-manager `Certificate` resources, prometheus-operator `PrometheusRule` resources, etc.

Would it be possible to add support for installing arbitrary Kubernetes resources so these things could be part of the same deployment?

This is a common pattern with Bitnami's Helm charts:

* an example of how to specify resources in the values.yaml: https://github.com/bitnami/charts/blob/d994badbdf500b707faf9a41caf00d085c8d3ed3/bitnami/nginx/values.yaml#L31
* an example of the template they use to generate the resources: https://github.com/bitnami/charts/blob/master/bitnami/nginx/templates/extra-list.yaml

colearendt:
close #115 

add helper to `rstudio-library` chart and implement in `rstudio-connect`, `rstudio-workbench`, and `rstudio-pm` charts

### DIFF
--- a/charts/rstudio-connect/Chart.lock
+++ b/charts/rstudio-connect/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: rstudio-library
   repository: file://../rstudio-library
-  version: 0.1.19
-digest: sha256:3a52152a14dd67c078b2b6b66747e31618aaa97abbe3e8255b59762766bbb463
-generated: "2021-12-17T12:28:47.58543-05:00"
+  version: 0.1.20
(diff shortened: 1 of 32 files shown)


atheriel:
This makes sense to me, but I'd suggest we call it `extraObjects` to match [Kubernetes terminology](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/) and the existing docstring (which uses "objects"). `extraDeploy` makes it seem like this is related to Deployments, which it is not.

atheriel (approved):

PR #129, opened 2022-01-03
add "pro" runtime.yaml

### LINKED ISSUE #92: Maybe add a "pro" option for `runtime.yaml` in the helm chart?
https://github.com/rstudio/connect/issues/20125

It would change our default from `false` to something else... like a magic string that changes behavior. Not sure if that is to be preferred or avoided...

Should also document more clearly how to run this with a separate file (and see how to do that in `helmfile`)

colearendt:
close #92 

Adding an "append runtime.yaml extras" is "possible" but pretty hairy with these file inclusions. I suspect we should hold off on that for now and see if changes to the product make `runtime.yaml` less required.


charts/rstudio-connect/templates/_helpers.tpl
```diff
@@ -89,7 +89,9 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 */}}
 {{- define "rstudio-connect.runtimeYaml" -}}
   {{- $runtimeYaml := .Values.launcher.customRuntimeYaml }}
-  {{- if $runtimeYaml }}
+  {{- if eq $runtimeYaml "pro" }}
+    {{- .Files.Get "default-runtime-pro.yaml" }}
+  {{- else if $runtimeYaml }}
```
aronatkins, line 94: I'm confused by the code compared to the documentation. This field is not really a Boolean, as documented. Here, we see that it might be a string-with-meaning (pro), or a string-with-meaning (YAML-as-string), or a YAML object, or some-value-that-is-not-truthy (default YAML).

All of this complexity is not really captured in the documentation.

What about:

`customRuntimeYaml` is an object allowing values such as:

* When not specified, use the base images (default).
* When falsy, use the base images (backwards-compatible).
* When "base", use the base images.
* When "pro", use the pro images.
* When a string, interpret as YAML.
* When an object, assume well-structured YAML.

It would also be handy to link to the "base" and "pro" YAML files as examples.

We could probably expand this field to take list values, but we can postpone that.

aronatkins (approved):
Approved, but some suggestions about improving the documentation.

aronatkins:
Should this bump the chart version?

charts/rstudio-connect/templates/_helpers.tpl
colearendt: I like that idea! Well said!

aronatkins (approved):

Issue #72, opened 2021-11-01
Improve dev documentation


colearendt:
- [ ] what version(s) of helm we test with / support
- [ ] what version of `ct` / how to install
- [ ] add testing for "minimum" helm version we support... basic `helm 3`?
- [ ] available just targets, dev dependencies, etc.
- [ ] `just` target for linting
- [ ] `just` target for regenerating RBAC like in CI
- [ ] what version of `helm-docs` / how to run locally

PR #133, opened 2022-01-04
Update RSPM version to 2021.12.0

glin:
This was my first time making any change to this repo, and I used https://github.com/rstudio/helm/pull/53 as an example. There were some seemingly-unrelated changes to the Values table in README.md when I regenerated that though. Are those changes okay, or should I revert them?


charts/rstudio-pm/README.md
```diff
@@ -142,3 +142,6 @@ The Helm `config` values are converted into the `rstudio-pm.gcfg` service config
 | strategy.type | string | `"RollingUpdate"` |  |
 | versionOverride | string | `""` | A Package Manager version to override the "tag" for the RStudio Package Manager image. Necessary until https://github.com/helm/helm/issues/8194 |
 
+----------------------------------------------
+Autogenerated from chart metadata using [helm-docs v1.5.0](https://github.com/norwoodj/helm-docs/releases/v1.5.0)
```
glin, line 146: Ohh, I see now that there's a GHA workflow to update the README. So you shouldn't have to regenerate the README manually?

If you do, it seems like I had a different version of helm-docs installed, so it would be nice to have that version documented somewhere, or in a Docker image perhaps.

colearendt: Yes! You shouldn't need to regenerate manually 😄 Although coming from forks it still fails.

But yes, I just added that to the list of things we need to improve in #72

colearendt (approved):
Looks awesome! Thanks Greg!

Issue #52, opened 2021-09-23
Ingress specification


pat-s:
Hi, thanks for the helm chart! 

I am having trouble getting the ingress right.

## Ingress specification

When I try to set the ingress in the current specification as

```yml
  - host: rsw.example.org
    paths:
      - path: /
```

I see

```
Error: failed to create resource: Ingress.extensions "rstudio-rstudio-workbench" is invalid: [spec.rules[0].http.paths[0].path: Invalid value: "map[path:/]": must be an absolute path, spec.rules[0].http.paths[0].backend: Required value: port name or number is required]
```

If I don't specify `paths` I get an error. Maybe there could be a default for this option?

```yml
  - host: rsw.example.org
    paths: []
```

```
 Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Ingress.spec.rules[0].http): missing required field "paths" in io.k8s.api.networking.v1beta1.HTTPIngressRuleValue
```

Guidance how to set the ingress properly would be highly appreciated! 

## pathType Option

Since k8s v1.18 it is common to also provide a `pathType` option in the ingress.
Would it be possible to adapt the template to support this?

See [the gitea helm chart](https://gitea.com/gitea/helm-chart/src/branch/master/templates/gitea/ingress.yaml) as an example.

So in the end I would like to write the following

```yml
ingress:
  enabled: true
  annotations: {}
  hosts:
    - host: rsw.example.com
      paths:
        - path: /
          pathType: Prefix

```

pat-s:
After many hours of trial and error I found the following specification which works

```yml
ingress:
  enabled: true
  hosts:
    - host: subdomain.example.org
      paths: [/]
```

[Reference](https://github.com/kubernetes/ingress-nginx/issues/4696#issuecomment-558062986)

This only works if `service.port` mapping is added in `values.yml` because it is referenced in 

https://github.com/rstudio/helm/blob/7caafdf5159f6c4e662bc0d5d740f1e588500b4a/charts/rstudio-workbench/templates/ingress.yaml#L3

and currently not defined in the `values.yml` template.

```yml
service:
  port: 8787
```

When missing it causes the following error

```
	│ Error: cannot patch "rsw" with kind Ingress: Ingress.extensions "rsw" is invalid: spec.rules[0].http.paths[0].backend: Required value: port name or number is required
```

pat-s:
After the latest ingress changes in #117, my previous ingress definition is failing.

[...]

PR #134, opened 2022-01-05
Allow new ingress pattern

### LINKED ISSUE #139: Issue with ingress in <=v1.19
Found by @dbkegley 

Our ingress apiVersion determination fails on >=1.14 and <=1.19 due to a whitespace issue

Tracking this down was difficult, although #105 could perhaps have helped. Helm retains an "internal" `KubeVersion` for use in templating - so we could potentially make use of that in CI to ensure templating works for these types of API / version checks

This also broke downstream comparisons so that other version gates did not work

colearendt:
add a helper to generate the path and pathType parts of ingress 

this allows a string to be passed for backwards compatibility, but will also
allow specifying the new pattern `path` and `pathType` keys

For #52 

close #139


atheriel (commented):
I think this workaround is sensible so that we avoid breaking existing configurations, but we should also make sure the older pattern is scrubbed from existing docs and values files (and appears only in tests). This could be a separate PR, though.

PR #135, opened 2022-01-06
add pod affinity definition

colearendt:
define the `pod.affinity` value as discussed [here](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity)
pass along to the pod deployment
bump version

(if we like this pattern, I will go ahead and add to the other charts as well)


aronatkins (approved):

aronatkins:
CC @dbkegley

PR #138, opened 2022-01-08
Update information on defining "profiles" in README

pat-s:
I believe the following information could be helpful to users as it is easy to get confused with the initial `profiles` key as it does not correspond to the `/etc/rstudio/profiles` file admins from non-k8s installations usually deal.

(Sorry for the whitespace changes, my editor strips them automatically - maybe you're fine with it given that these seem to be superfluous).

Also I was wondering if an additional information stating that "defining `session-timeout-minutes` and `session-timeout-kill-hours` is really important for k8s deployments" might be helpful.
Otherwise sessions will never get suspended and hence the underlying resources will always stay up, possibly incurring high costs which could be easily avoided.


colearendt:
Thanks for the useful doc update! I batched a few other little issues together since we are bumping the version anyways 😄

PR #140, opened 2022-01-11
Fix rsw ha

### LINKED ISSUE #137: RStudio Workbench Load Balancing
The [readme](https://github.com/rstudio/helm/blob/f73fdd04cecfaead6d48f230e2ae722728448ed4/charts/rstudio-workbench/README.md?plain=1#L53) seems to indicate that setting replicas > 1 should be enough to enable load balancing as of the 2021.09 release.

However, with replicas > 1 I'm running into some authentication issues:

```
2022-01-08T04:37:43.467818Z [rserver] ERROR system error 74 (Bad message) [description: error:0407109F:rsa routines:RSA_padding_check_PKCS1_type_2:pkcs decoding error, description: Failed sign-in - unable to decrypt password - error]; OCCURRED AT rstudio::core::Error rstudio::core::system::crypto::rsaPrivateDecrypt(const string&, std::__cxx11::string*) src/cpp/core/system/Crypto.cpp:377; LOGGED FROM: void rstudio::server::pam_auth::{anonymous}::doSignIn(const rstudio::core::http::Request&, rstudio::core::http::Response*) src/cpp/server/ServerPAMAuth.cpp:134
```

When `replicas=1` I'm able to log in successfully.

it looks like the load-balancer file referenced in the [documentation](https://docs.rstudio.com/ide/server-pro/load_balancing/configuration.html#defining-nodes) isn't being created. The readme at `/etc/rstudio/README.md` in the container suggests that it should be located at `/mnt/load-balancer/rstudio/load-balancer`.

Is there a setting that I'm missing?

colearendt:
pass information to the workbench pod so that it can create the load-balancer file

we omitted this (big woops) on the shift during Ghost Orchid, and forgot to touch the file
this means that workbench has been functioning in a "clusterless" capacity for the most-part

close #137

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-workbench
 description: Official Helm chart for RStudio Workbench
-version: 0.5.3
+version: 0.5.4
 apiVersion: v2
 appVersion: 2021.09.0-351.pro6
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
--- a/charts/rstudio-workbench/NEWS.md
+++ b/charts/rstudio-workbench/NEWS.md
@@ -1,3 +1,10 @@
+# 0.5.4
+
+- BUGFIX: address an important issue in RStudio Workbench load balancing
+  - Ever since 0.5.0, we did not create a `load-balancer` file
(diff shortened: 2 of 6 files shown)


colearendt:
Tested in our demo environment. All seems to be well. We will keep an eye out for whether a more robust fix with IP / etc. is necessary until we can improve the product.

PR #154, opened 2022-02-15
Add other charts

colearendt:
Add the `prepull-daemonset` chart, which could be useful for customers

Namespaced separately (in `other-charts`) in sources here to make clear that the lifecycle / support level is different

TODO:
- [x] ensure chart releaser will publish these
- [x] ensure that releaser rebuild script will not be broken by a separate directory
- [x] add explanatory section to the README


dbkegley (approved):
LGTM!

other-charts/prepull-daemonset/templates/daemonset.yaml
```diff
@@ -0,0 +1,34 @@
...
+        rollme: {{ randAlphaNum 10 | quote }}
+        {{- end }}
+      labels:
+        name: {{ .Chart.Name | trunc 63 | trimSuffix "-" }}
+    spec:
+      initContainers:
+      {{- range .Values.images }}
+        - name: prepull-{{ .name }}
+          image: {{ .image }}
+          imagePullPolicy: Always
```
dbkegley, line 28: Not necessarily in this PR, but I think some customers will need to set an  `imagePullSecrets` here from an optional `{{ .imagePullSecret }}` in the config

other-charts/prepull-daemonset/README.md.gotmpl
```diff
@@ -0,0 +1,36 @@
...
+## Upgrade
+
+To run the pre-pull again, there are a few options:
+```
+# by default, every upgrade is randomized / does a rollout (`--set randomize=false` to disable this behavior)
+helm upgrade install preepull-daemonset rstudio/prepull-daemonset
+
+# or use kubectl
+kubectl rollout restart daemonset/prepull-daemonset
+```
```
dbkegley, line 32: What are the steps a user would need to follow when an new image becomes available? Should they run `helm upgrade` or `kubectl rollout` on a cron to fetch the latest on some interval or is there a better way to make sure the cache stays warm?

dbkegley: One option for this might be to use a `sleep` for a configurable interval in the `pause` container. Since the daemonset defaults to `restartPolicy: Always` then it should just restart and pull the images again

dbkegley (approved):

other-charts/prepull-daemonset/templates/daemonset.yaml
colearendt: Great call! I'll create an issue for follow-up

other-charts/prepull-daemonset/README.md.gotmpl
colearendt: That's a great call. To date we have ignored that question. If you `helm upgrade` (even with same values) the `randomize: true` setting will automatically force a re-pull. However, a way to "keep containers pulled" could be useful.

One downside of doing that is that kubernetes tries to ensure that disks do not get overwhelmed. If you put too many images into the pull or if lots of other images are in use, you could end up overwhelming your disks and cause a bunch of flapping/turnover in the image cache. I'll create an issue to follow-up!

dbkegley: Whichever direction we decide to go, I think just an short note in the readme about how/when to run the rollout would be useful. It wasn't immediately clear to me when/ why I would need to run `helm upgrade`

PR #174, opened 2022-04-06
Add default mode

### LINKED ISSUE #151: Specify `defaultMode` for session-scoped config files
Since the session-scoped configuration files ([`config.session.*`](https://github.com/rstudio/helm/blob/920b136037e29b7ec846a0d3021591eead497c0f/charts/rstudio-workbench/values.yaml#L279) in the `values.yaml`) are mounted on both the server and the sessions, we are using them to create an `entrypoint.sh` file on the sessions in order to override the entrypoint for sessions (background on the entrypoint requirements are [here](https://support.rstudio.com/hc/en-us/articles/360059743753-Customizing-the-Entrypoint-with-RStudio-Workbench-RStudio-Server-Pro-Job-Launcher-and-Kubernetes), thanks @colearendt):

```
config:
  session:
    entrypoint.sh: |
      #!/bin/bash

      set -xe

      # custom stuff here

      # Launch the session based on RSTUDIO_TYPE
      if [[ "$RSTUDIO_TYPE" == "session" ]]; then
        echo 'Starting RStudio session...'
        exec "/usr/lib/rstudio-server/bin/rserver-launcher" "$@"
      elif [[ "$RSTUDIO_TYPE" == "adhoc" ]]; then
        echo 'Starting RStudio ad hoc job...'
        /bin/bash "$@"
      else
        echo "ERROR: Unknown RSTUDIO_TYPE: $RSTUDIO_TYPE"
        exit 1
      fi
```

This mounts correctly in the session at `/mnt/session-configmap/rstudio/entrypoint.sh`, however it can't be used directly as the entrypoint as part of a [`config.profiles.*.*.job-json-overrides`](https://github.com/rstudio/helm/blob/920b136037e29b7ec846a0d3021591eead497c0f/charts/rstudio-workbench/values.yaml#L338) entry because it's not executable. To get around this, we've had to hack together a workaround to copy the script to a new location, make it executable, and call it:

```
config:
  profiles:
    launcher.kubernetes.profiles.conf:
      "*":
          - target: /spec/template/spec/containers/0/command
            json:
            - sh
            - -c
            - 'cp /mnt/session-configmap/rstudio/entrypoint.sh /bin/entrypoint.sh && chmod +x /bin/entrypoint.sh && /bin/entrypoint.sh $@'
            - sh
            name: commandOverride   
```

It would be helpful if we could pass in a [`defaultMode`](https://kubernetes.io/docs/concepts/configuration/secret/#secret-files-permissions) with [`config.session.*`](https://github.com/rstudio/helm/blob/920b136037e29b7ec846a0d3021591eead497c0f/charts/rstudio-workbench/values.yaml#L279) to mount this file as executable from the start, which would allow us to simplify the `job-json-override` to:

```
config:
  profiles:
    launcher.kubernetes.profiles.conf:
      "*":
          - target: /spec/template/spec/containers/0/command/0
            json: /mnt/session-configmap/rstudio/entrypoint.sh
            name: commandOverride   
```

This might be useful for session-scoped secrets and server-scoped configs and secrets as well, but our particular use-case is just the session-scoped config files.

colearendt:
this sets most (all?) options for defaultMode on configmap and secret volumes in the `rstudio-workbench` chart
using octal numbers in helm is a bit strange, but seems to work correctly

close #151

### DIFF
--- a/charts/rstudio-workbench/README.md
+++ b/charts/rstudio-workbench/README.md
@@ -1,6 +1,6 @@
 # RStudio Workbench
 
-![Version: 0.5.12](https://img.shields.io/badge/Version-0.5.12-informational?style=flat-square) ![AppVersion: 2022.02.1-461.pro1](https://img.shields.io/badge/AppVersion-2022.02.1--461.pro1-informational?style=flat-square)
+![Version: 0.5.13](https://img.shields.io/badge/Version-0.5.13-informational?style=flat-square) ![AppVersion: 2022.02.1-461.pro1](https://img.shields.io/badge/AppVersion-2022.02.1--461.pro1-informational?style=flat-square)
 
 #### _Official Helm chart for RStudio Workbench_
 
@@ -23,11 +23,11 @@ As a result, please:
 
 ## Installing the Chart
 
-To install the chart with the release name `my-release` at version 0.5.12:
+To install the chart with the release name `my-release` at version 0.5.13:
 
 ```bash
 helm repo add rstudio https://helm.rstudio.com
-helm install my-release rstudio/rstudio-workbench --version=0.5.12
+helm install my-release rstudio/rstudio-workbench --version=0.5.13
 ```
 
 ## Required Configuration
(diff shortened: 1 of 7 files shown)


atheriel:
From #151 I see why you might want to change the mode for some configuration files, but I'm not sure I understand why we'd want to expose everything -- it might be worth expanding on the use cases in the NEWS item to help explain them.

charts/rstudio-workbench/README.md
```diff
@@ -323,6 +323,7 @@ config:
 | affinity | object | `{}` |  |
 | args | list | `[]` | args is the pod container's run arguments. |
 | command | list | `[]` | command is the pod container's run command. By default, it uses the container's default. However, the chart expects a container using `supervisord` for startup |
+| config.defaultMode | object | `{"jobJsonOverrides":420,"pam":420,"prestart":493,"secret":384,"server":420,"session":420,"sessionSecret":272,"startup":493,"userProvisioning":384}` | defaultMode used for mounting the various configuration configmaps |
```
atheriel, line 326: The generated documentation here is really poor -- "object" and numbers that have been translated from Octal -- so maybe we should try to improve it by having per-key docstrings or store the modes as strings.

colearendt:
Yeah that makes sense - I will try to explain in NEWS. Basically I was thinking about targeting consistency / avoiding a onesie-twosie PR path where we keep having to add flexibility once someone has a use case we didn't think about 😅 It can also be a good escape hatch with weird mounting shenanigans (like we have hit lately with RSPM) 😅

charts/rstudio-workbench/values.yaml
```diff
@@ -279,6 +279,36 @@ global:
   secureCookieKey: ""
 
 config:
+  # -- defaultMode used for mounting the various configuration configmaps
```
atheriel, line 282: This is still generating a weird docstring; I'd suggest removing this line.

atheriel (approved):

Issue #35, opened 2021-08-18
Set up a values.schema.json


colearendt:
The `values.schema.json` file enables some snazzy ArtifactHub behavior that allows users to explore valid values ([on display here](https://artifacthub.io/packages/helm/artifact-hub/artifact-hub))

This is also a part of helm's validation behavior, so can protect users from some invalid values: https://helm.sh/docs/topics/charts/#schema-files

An example that renders the artifacthub behavior above: https://gist.github.com/colearendt/7e5066c92d74df43010e16e3c118938c

- [ ] Figure out a way to define JSON schema in a sane way (perhaps related to #17 ). Is there a UI explorer that makes it easy to interact with these types of files?
- [ ] A way to generate `values.yaml` from the JSON schema?
- [ ] Embed the values.schema.json file into the chart, and release it
- [ ] Ensure that the display is appropriate on artifacthub

Issue #17, opened 2021-08-05
Ensure consistency in helm values across charts


colearendt:
We should have a portion of CI that ensures we are aware when we add new values

(Not sure if this is do-able via CI...)

- detect if similar values are used? (i.e. `serviceAccountName` vs. `serviceAccount.name`)
- aggregate values keys across charts
- aggregate default values across charts (for shared keys)

PR #176, opened 2022-04-07
Allow Launcher to self-configure Kubernetes

bschwedler:
Now that Launcher supports self-configuring Kubernetes, we no longer have to dynamically generate the `launcher.kubernetes.conf` file.

In addition, these changes also support multiple `[cluster]` sections in the `launcher.conf` file.


charts/rstudio-library/templates/_config.tpl
```diff
@@ -50,18 +50,20 @@
   {{- $keys | nindent 2 }}
 {{- else }}
 {{- range $parent, $child := $keys -}}
-  {{- if kindIs "map" $child }}
-    {{- if not ( kindIs "slice" $keys ) -}}
+  {{/* ini files may have multiple sections with the same name */}}
+  {{- $sections := ( (kindIs "slice" $child) | ternary $child ( list $child ))}}
```
bschwedler, line 54: Supporting multiple `[cluster]` sections supports using different config files and scratch paths for launcher plugins.

If these configurations specify different values for `kubernetes-namespace` and RBAC is not configured correctly, RSW may fail to start up.

bschwedler:
@colearendt This is now ready for review.

bschwedler:
This also addresses #180

charts/rstudio-workbench/prestart-launcher.bash
```diff
@@ -8,22 +8,18 @@ main() {
...
   sa_token="$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)"
 
-  _logf 'Loading service account ca.crt'
-  local ca_string
-  ca_string="$(tr -d '\n' <"${cacert}" | base64 | tr -d '\n')"
-
   _logf 'Ensuring %s exists' "${dyn_dir}"
   mkdir -p "${dyn_dir}"
 
+  # Empty if enabled, set to "disabled"
```
colearendt, line 19:
```suggestion
  # Empty if enabled, set to "disabled" by default
```

colearendt (commented):
This looks great! I want to interactively play with the library changes a smidge (which I plan to do here shortly), but wanted to send this review across.

Would you mind adding a few simple NEWS entries for the RSC / RSW / RSPM charts with a "bumped library chart version" note? Also a NEWS entry for the rstudio-library chart highlighting what we changed would be good too! A bit tedious unfortunately, but is helpful to trace changes to different chart versions 😄

bschwedler:
@colearendt Have you had a chance to play with these changes yet?

If they look good, I will go ahead and update the chart in this PR.

colearendt:
Resolved merge conflicts and fixed a little bug in the default values that made templating fail (when `.serviceAccountCreate` is empty (the default)). I think this should be good to go! Just need to sanity check that all the version numbers match - will try to do that later tonight

colearendt (approved):
OK! I think this is good from my point of view. My first pass was a bit naive and (if I recall correctly) introduced the same problem we had before with a tautology `default` 🙈 

The problem with `false` and `interface {}` both being "falsey" / "empty" was a tricky one. I think comparing the type is probably our better bet - would love your thoughts before merging though!

charts/rstudio-library/templates/_rbac.tpl
```diff
@@ -12,14 +12,14 @@
 */ -}}
 {{- define "rstudio-library.rbac" -}}
 {{- $serviceAccountAnnotations := default (dict) .serviceAccountAnnotations }}
-{{- $serviceAccountCreate := default true .serviceAccountCreate }}
+{{- $serviceAccountCreate := or ( and (not (kindIs "bool" .serviceAccountCreate)) (empty .serviceAccountCreate) ) (not (not .serviceAccountCreate)) }}
```
bschwedler, line 15: It like to take another look/whack at this. If we can do some boolean algebra to invert portions of this logic I think it could me more understandable.

My logic missed the case where it was unset and we wanted to create the service account.

bschwedler: After our conversation I'd prefer that we implement this as a type check as in https://github.com/rstudio/helm/blob/main/charts/rstudio-library/templates/_debug.tpl

Or we could move all values validation to a JSON schema: https://github.com/rstudio/helm/issues/35

charts/rstudio-library/NEWS.md
```diff
@@ -1,3 +1,21 @@
...
+        launcher.conf:
+          cluster:
+            - name: Cluster1
+              type: Kubernetes
+            - name: Cluster2
+              type: Kubernetes
+              config-file: /path/to/config/file
+    ```
+
+- Fixes issue where RBAC service account is always created even when `serviceAccount.create: false`
```
colearendt, line 17: Maybe make a note here that this now requires `serviceAccount.create` to be provided as a boolean (and not empty)?

charts/rstudio-library/templates/_rbac.tpl
```diff
@@ -12,7 +12,7 @@
 */ -}}
 {{- define "rstudio-library.rbac" -}}
 {{- $serviceAccountAnnotations := default (dict) .serviceAccountAnnotations }}
-{{- $serviceAccountCreate := default true .serviceAccountCreate }}
+{{- $serviceAccountCreate := eq .serviceAccountCreate false | ternary false true }}
```
colearendt, line 15: Ah interesting- I must admit that I'm surprised we are not making the type check here. I guess it allows us to have a cleaner error message if the downstream charts do the error check though.

Would you mind making a note in the comment above that this is required to be a boolean and will fail otherwise?

colearendt (approved):
LGTM with just a few caveats! Is `rstudio-launcher-rbac` the only place where we use the library chart without providing "create" as a boolean? i.e. are type checks needed elsewhere?

charts/rstudio-library/templates/_rbac.tpl
bschwedler: Ah. I misunderstood how this was being done in the library.
It's somewhat of a dependency graph issue because the `serviceAccountCreate` value is actually set by the various charts that consume the library and passed to the named template.

Each of the charts consuming the library set this with a different values file entry: https://github.com/rstudio/helm/search?q=serviceAccountCreate&type=code

It might be better to move this validation into the library chart, and handle this better in #35 and #17

bschwedler: This has been moved into the library chart.

colearendt (approved):
Love it!! Exciting stuff!

PR #181, opened 2022-04-19
Update for RStudio Package Manager 2022.04.0-7

shepherdjerred:
Update RStudio Package Manager to `2022.04.0-7` which is the newest released version.


colearendt (approved):
LGTM, thanks!

PR #183, opened 2022-04-20
Added the postStart hook feature

rodrigobrim:
### Problem

We need to serve the rstudio-pm Helm release via IaC, but I can't find a way to do so.

### Proposed solution

Add the postStart lifecycle hook into the deployment template, allowing the execution of arbitrary commands.

### Values example

```
...
license:
  file:
    mountPath: /etc/rstudio-licensing
    secret: rspm-license
    secretKey: Data
  server: false
lifecycle:
    postStart:
      exec:
        command:
        - /bin/bash
        - -c
        - |
          sleep 3 &&
          rspm create repo --name=prod-cran --description='Access CRAN packages' &&
          rspm subscribe --repo=prod-cran --source=cran &&
          rspm sync --type=cran& exit
...
```

### DIFF
--- a/charts/rstudio-pm/Chart.yaml
+++ b/charts/rstudio-pm/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-pm
 description: Official Helm chart for RStudio Package Manager
-version: 0.3.4
+version: 0.3.5
 apiVersion: v2
 appVersion: 2021.12.0-3
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
--- a/charts/rstudio-pm/Chart.yaml
+++ b/charts/rstudio-pm/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-pm
 description: Official Helm chart for RStudio Package Manager
-version: 0.3.5
+version: 0.3.6
 apiVersion: v2
 appVersion: 2022.04.0-7
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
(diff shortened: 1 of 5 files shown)


colearendt (changes requested):
Thanks for the PR @rodrigobrim ! 

A couple of thoughts:
- One way to do this is to run a sidecar container that provisions state in the RSPM cluster and then sleeps forever. This may be undesirable though.
- Are you going to do this on every container startup? I'm curious here - will reach out offline
- I am not opposed to adding this, or `lifecycle` generally. Honestly, I feel like the whole `lifecycle` parameter may be a preferred item to expose so that other lifecycle hooks are available as well! Would you mind changing scope to that?
- If you can add a `NEWS.md` file entry, that would be great too!


You will also need to rebase / pull from the latest main, because RSPM just upgraded and the chart version changed 😅

rodrigobrim:
> Thanks for the PR @rodrigobrim !
> 
> A couple of thoughts:
> 
> * One way to do this is to run a sidecar container that provisions state in the RSPM cluster and then sleeps forever. This may be undesirable though.
> * Are you going to do this on every container startup? I'm curious here - will reach out offline
> * I am not opposed to adding this, or `lifecycle` generally. Honestly, I feel like the whole `lifecycle` parameter may be a preferred item to expose so that other lifecycle hooks are available as well! Would you mind changing scope to that?
> * If you can add a `NEWS.md` file entry, that would be great too!
> 
> You will also need to rebase / pull from the latest main, because RSPM just upgraded and the chart version changed sweat_smile

Sure. I agree with the `generally lifecycle`. I'll pull e rewrite it.

About the sidecar, how do you imagine making it using the Helm release deployment?

colearendt:
Ahh yes, I forgot that you have to use the `rspm` CLI for this (not a REST API) 😞 So the sidecar probably wouldn't work in this case without some pretty heavy lifting. Never-mind haha 😄 

Would you mind adding the `NEWS.md` entry describing your change?

rodrigobrim:
Exactly. We have a lot of possibilities, but I believe it's the simplest and fastest. After that, we can discuss the next steps with better strategies.

NEWS.md updated.

PR #184, opened 2022-04-22
Update Connect to 2022.04.0

joshfrench:

PR #185, opened 2022-04-22
Update CI defaults for connect 

dbkegley:
Depends on https://github.com/rstudio/helm/pull/184

We were missing `podAffinityTerm` in our ci/complex-values for connect which was causing the install step to fail. The sidecar was also entering a crashloop backoff because it exits immediately.

I think we still need to provide a `license.key` for connect for this part of CI to work properly but I'm not sure how to do that

### DIFF
--- a/charts/rstudio-connect/Chart.yaml
+++ b/charts/rstudio-connect/Chart.yaml
@@ -1,8 +1,8 @@
 name: rstudio-connect
 description: Official Helm chart for RStudio Connect
-version: 0.2.26
+version: 0.2.27
 apiVersion: v2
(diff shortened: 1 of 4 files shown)


colearendt (approved):
LGTM!

colearendt:
Merp. My memory is terrible. This is defunct right?

dbkegley:
Merged main if we still want these updates!

colearendt (approved):
I like it! Much appreciated sir!

colearendt:
Oops - just need to bump chart version 🙈

PR #189, opened 2022-04-29
Add sidecarContainers value to rstudio-pm template

rodrigobrim:
Add this feature to allow pod sidecar declaration.

### DIFF
--- a/charts/rstudio-pm/templates/deployment.yaml
+++ b/charts/rstudio-pm/templates/deployment.yaml
@@ -124,6 +124,9 @@ spec:
             {{- toYaml . | nindent 10 }}
           {{- end }}
         {{- end }}
+      {{- if .Values.sidecarContainers }}
+      {{- toYaml .Values.sidecarContainers | nindent 6 }}
+      {{- end }}
       volumes:
       - name: rstudio-pm-config
         configMap:
--- a/charts/rstudio-pm/values.yaml
+++ b/charts/rstudio-pm/values.yaml
@@ -191,3 +191,6 @@ config:
   Launcher:
     ServerUser: root
     AdminGroup: root
+
+# -- sidecar container list
+sidecarContainers: []
--- a/charts/rstudio-pm/Chart.yaml
+++ b/charts/rstudio-pm/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-pm
 description: Official Helm chart for RStudio Package Manager
-version: 0.3.6
+version: 0.3.7
 apiVersion: v2
 appVersion: 2022.04.0-7
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
--- a/charts/rstudio-pm/NEWS.md
+++ b/charts/rstudio-pm/NEWS.md
@@ -1,3 +1,7 @@
+# 0.3.7
+
+-  Add `sidecarContainers` value. This allows adding a list of additional containers.
+  
 # 0.3.6
 
 -  Add `pod.lifecycle` value. This allows running lifecycle hooks like postStart commands!


atheriel:
* I have a feeling `extraContainers` is more in line with Helm community naming conventions. It's used by Grafana and Ingress-NGINX, for example.

* What is the intended use case for this, if you don't mind sharing?

atheriel:
Another excellent reason for not using `sidecarContainers`: Kubernetes has [not actually accepted](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/753-sidecar-containers) the proposed sidecar container APIs at the time of writing, so let's avoid it and use more future-proof terminology.

charts/rstudio-pm/values.yaml
```diff
@@ -191,3 +191,6 @@ config:
   Launcher:
     ServerUser: root
     AdminGroup: root
+
+# -- sidecar container list
+sidecarContainers: []
```
atheriel, line 196: Please use `extraContainers` instead, and update existing references. I've also added suggested documentation, from looking at a few other sources:

```suggestion
# -- Additional containers to be added to the pod.
extraContainers: []
```

atheriel (changes requested):

rodrigobrim:
> * I have a feeling `extraContainers` is more in line with Helm community naming conventions. It's used by Grafana and Ingress-NGINX, for example.
> * What is the intended use case for this, if you don't mind sharing?

Changed from sidecarContainers to extraContainers.

charts/rstudio-pm/values.yaml
```diff
@@ -191,3 +191,6 @@ config:
   Launcher:
     ServerUser: root
     AdminGroup: root
+
+# -- sidecar container list
+extraContainers: []
```
colearendt, line 196: Missing terminating newline character

charts/rstudio-pm/templates/deployment.yaml
```diff
@@ -124,6 +124,9 @@ spec:
             {{- toYaml . | nindent 10 }}
           {{- end }}
         {{- end }}
+      {{- if .Values.extraContainers }}
+      {{- toYaml .Values.extraContainers | nindent 6 }}
```
colearendt, line 128: A bit of a nit, but indenting this one smidge further can make it easier to read the control flow

@atheriel do you have thoughts on whether to treat this like a template or not?

colearendt (commented):
Agree with Aaron's feedback!

charts/rstudio-pm/templates/deployment.yaml
atheriel: Convention seems to be not to, I agree. Is it problematic to add later if requested?

colearendt:
Nah, I don’t think so. I think leaving off for now seems best

charts/rstudio-pm/templates/deployment.yaml
rodrigobrim: Hey guys. Do I need to fix anything here?

colearendt: A bit of a nit, but if you could indent this line two more characters, that would be appreciated! Other than that, LGTM!

rodrigobrim: Please check if the latest push is like you requested.

colearendt (approved):
LGTM!

I manually ran `just docs` to get the docs up to date - in the future if you can run `just docs` yourself (will require a bit of setup - apologies docs are not better on that (#72)) or create a branch on the helm repo itself, that will make the CI story happier 😄 (reach out if you do not have access and we can figure that out)

rodrigobrim:
> LGTM!
> 
> I manually ran `just docs` to get the docs up to date - in the future if you can run `just docs` yourself (will require a bit of setup - apologies docs are not better on that (#72)) or create a branch on the helm repo itself, that will make the CI story happier smile (reach out if you do not have access and we can figure that out)

I don't have access ;-(

atheriel (approved):
@colearendt I think we should merge this and then run the doc update command separately. IIUC, there is a permissions issue because this PR originates from a fork.

PR #196, opened 2022-05-10
enable pod-level configuration of securityContext for rstudio-connect

scott-grimes:


colearendt:
Thanks for the PR! Would you mind moving this under the `pod:` key, perhaps? I'm not sure what convention is here, but it is nice having these things grouped together IMO (although I can see how putting the `securityContext`s together is nice too 🙈 

https://github.com/rstudio/helm/blob/ac2ad2b2e3c84309a456f6229934e5170c82164e/charts/rstudio-connect/values.yaml#L71

PR #198, opened 2022-05-12
Enable securityContext for pod and container

rodrigobrim:
As part of our plan to run the container as an ordinary user, we need to change the default `container[0].securityContext.privileged` from true to false.

We also need the ability to change the pod security context, allowing the `fsGroup` storage parameter.

### DIFF
--- a/charts/rstudio-pm/values.yaml
+++ b/charts/rstudio-pm/values.yaml
@@ -88,8 +88,9 @@ rstudioPMKey: false
 command: false
 # -- args is the pod's run arguments. By default, it uses the container's default
 args: false
-# -- pod [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) defines privilege and access control settings
-securityContext: {}
+# -- container [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) defines privilege and access control settings
+securityContext:
+  privileged: true
 
 pod:
   # -- env is an array of maps that is injected as-is into the "env:" component of the pod.container spec
@@ -104,9 +105,8 @@ pod:
   annotations: {}
   # -- Container [lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/)
   lifecycle: {}
-  # -- container [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) defines privilege and access control settings
-  securityContext:
-    privileged: true
+  # -- pod [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) defines privilege and access control settings
+  securityContext: {}
 
 # -- Extra objects to deploy (value evaluated as a template)
 extraObjects: []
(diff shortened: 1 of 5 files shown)


jonyoder (approved):
Looks good to me. It would probably be wise to have someone more familiar with helm (and more involved with this repo) to also take a look.

rodrigobrim:
> Looks good to me. It would probably be wise to have someone more familiar with helm (and more involved with this repo) to also take a look.

Sure. Aaron and Cole added.

colearendt (changes requested):
Looks good! I didn't realize we still run this as privileged. A couple of things:

- it looks like you mixed up the convention with the other products. You have `pod.securityContext` setting the context on the container, and `securityContext` setting the context on the pod? I will need to take a deeper look at the other charts, but I think this is mixed up here (even the docs?)

- Also, if we are keeping the default as `privileged: true`, we should put that in the default values, rather than set the default in an `else` in the helm chart

rodrigobrim:
> Looks good! I didn't realize we still run this as privileged. A couple of things:
> 
> * it looks like you mixed up the convention with the other products. You have `pod.securityContext` setting the context on the container, and `securityContext` setting the context on the pod? I will need to take a deeper look at the other charts, but I think this is mixed up here (even the docs?)
> * Also, if we are keeping the default as `privileged: true`, we should put that in the default values, rather than set the default in an `else` in the helm chart

They are not totally mixed. In the pod.securityContext there's no privileged config. In the container.securityContext, there's no fsGroup config. They are independent in some aspects.

About the if, I'll set it in the values.

colearendt:
Oh no, my point is that you are using `pod.securityContext` as "container security context" and `securityContext` as "pod security context." Those should be reversed 😄 

Expected:
- `securityContext`: container security context
- `pod.securityContext`: pod security context

charts/rstudio-pm/values.yaml
```diff
@@ -88,6 +88,9 @@ rstudioPMKey: false
 command: false
 # -- args is the pod's run arguments. By default, it uses the container's default
 args: false
+# -- container [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) defines privilege and access control settings
+securityContext:
```
atheriel, line 92: I've looked at [various](https://github.com/kubernetes/ingress-nginx/blob/24b8e32b088bec15e1f40deab8fefc8bd1b6c2da/charts/ingress-nginx/values.yaml#L825) [community](https://github.com/grafana/helm-charts/blob/f942ba605f8780518b6ac7f52eb68fb275c02e8b/charts/grafana/values.yaml#L101) charts and the convention is to use `containerSecurityContext` for this to disambiguate. Let's follow that convention.

atheriel (changes requested):

atheriel (approved):
This looks fine to me now. @colearendt do you have an opinion on whether `containerSecurityContext` should be top-level or nested under `pod`? Are these changes in line with what you expected?

colearendt (approved):
I do like that convention within `pod`, I think, and consistency w/ the community is a win! We should take a follow-up issue to move our other charts towards this pattern.

PR #199, opened 2022-05-13
Set up baseline for launcher templating

### LINKED ISSUE #215: rbac: pods/log actions
My k8s environment does not currently allow  `Verbs:["create" "update" "patch" "delete"]` for `Resources:["pods/log"]`

Any insight as to why these permissions are needed?

Could "pods/log" be limited to ["get" "watch" "list"]?

e.g.
`Resources:["pods/log"]` -> `Verbs:["get" "watch" "list"]`
`Resources:["pods", "pods/attach", "pods/exec"]` -> `Verbs:["get" "watch" "list" "create" "update" "patch" "delete"]`

colearendt:
**EDIT:** (see revision history for details on previous approach which involved base64 encoded files 🤢 )

Instead of the previously discussed approach, we have decided to:
- put template helpers into the library chart
- store copies of the launcher templates (versioned) and helm templates in the `examples/` directory (For easy "diffing")
- embed the needed versions of the templates (per chart) into the app chart itself (i.e. RSC, RSW, etc.)
- fix #215 by relaxing requirements for `pod/logs`


sagerb:
After a few readings, I think I understand what you've done, although I'm a bit confused looking at the different files in their directory locations and trying to determine if you've extended the product templates yet or not.

I'd like to suggest that we look to the areas which you'd like to extend and then go forward with implementing those generic extensions in the base product templates (seem useful). I like your approach of including another file which would then have static custom data available to the template (that gets around my need to have more and more info passed through the product just to be available within the template).

Putting my confusion mentioned above aside, I'm excited to start prototyping with this in place. I should be able to focus on that early next week. :-)

colearendt:
Updated this PR to be just the examples, so we can have different approaches in different branches

bschwedler (commented):
I would like to see us resolve this by pulling support for at least labels and annotations in Launcher itself.  There is still some design and implementation details to work out, but I think it will be a better path forward than having to manage the included templates.

https://github.com/rstudio/launcher/issues/250

PR #205, opened 2022-05-18
DRAFT: Tpl separate files

colearendt:
First pass at #199  - an attempt at using:
- separate files per inclusion
- values from helm users injected into the separate file inclusions
- a base64 trick for being able to modify yaml directly and then embed into helper functions (`$.Files.get` does not allow referencing files in the library chart)


colearendt:
This pattern ultimately proved to be pretty tedious to use. Instead, we are planning to use a "values" approach by parsing JSON and providing to the template:
image
image

This approach is displayed in #216 #213 and #209

Issue #207, opened 2022-05-24
rstudio-connect exporter sidecar resources not configurable


scott-grimes:
Sidecar containers should have their resources configurable, many cluster administrators enforce the presence of resource limits/requests to prevent scheduling problems

PR #208, opened 2022-05-24
sidecar resources configurable

scott-grimes:
allows configuration of the sidecars resources. also adopt the 

```
{{- with .Values.resources }}
  {{ toYaml . | nindent x }}
{{- end }}
```

pattern for resource definitions, which gives more flexibility to users and is more concise. 

fixes https://github.com/rstudio/helm/issues/207


colearendt (approved):
This looks awesome! Thanks for the PR!

I hope you don't mind - took the liberty to bump version / add NEWS / regenerate docs. It's a bit tedious that our CI for this does not work well from forks

PR #209, opened 2022-05-25
DRAFT: add basic tpl data parsing attempt

colearendt:
Another draft (related to #205 ) but exploring a "parsing a string into data" pattern for job customization

### DIFF
--- a/charts/rstudio-library/example-template-data.yaml
+++ b/charts/rstudio-library/example-template-data.yaml
@@ -6,3 +6,4 @@ pod:
   initContainers: []
   imagePullPolicy: ""
   imagePullSecrets: []
+  serviceAccountName: ""
--- a/charts/rstudio-library/example-template-data.yaml
+++ b/charts/rstudio-library/example-template-data.yaml
@@ -1,0 +1,8 @@
+service:
+  type: NodePort
+pod:
+  volumes: []
+  volumeMounts: []
+  initContainers: []
+  imagePullPolicy: ""
+  imagePullSecrets: []
--- a/charts/rstudio-library/templates/_launcher-templates.tpl
+++ b/charts/rstudio-library/templates/_launcher-templates.tpl
@@ -1,0 +1,34 @@
+{{- define "rstudio-library.templates.job" -}}
+  {{- include "rstudio-library.templates.b64.base.job" nil | b64dec }}
+{{- end }}
+
+{{- define "rstudio-library.templates.service" }}
+  {{- include "rstudio-library.templates.b64.base.service" nil | b64dec }}
+{{- end }}
+
+{{- define "rstudio-library.templates.configmap" -}}
+job.tpl: |
+{{- include (printf "rstudio-library.templates.b64.%s.job" . ) nil | b64dec | nindent 2 }}
+service.tpl: |
+{{- include (printf "rstudio-library.templates.b64.%s.service" . ) nil | b64dec | nindent 2 }}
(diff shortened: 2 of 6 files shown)


charts/rstudio-library/files/helm-2/service.tpl
```diff
@@ -20,4 +21,4 @@ spec:
   selector:
     job-name: {{toYaml .Job.id }}
   clusterIP: ''
-  type: NodePort
+  type: {{ $templateData.service.type }}
```
sagerb, line 24: Since you are conditional based on the presence of volume mounts, should you be conditional here (and if they don't provide it, then provide a default)?

charts/rstudio-library/files/helm/service.tpl
```diff
@@ -20,4 +21,4 @@ spec:
   selector:
     job-name: {{toYaml .Job.id }}
   clusterIP: ''
-  type: NodePort
+  type: {{ $templateData.service.type }}
```
sagerb, line 24: same comment as before applies here - conditional with a default?

sagerb (commented):
This looks good, although you could quickly imagine the template becoming unmaintainable if each and every k8s setting was conditionally checked for within the template. 

This also is using the templates to set global settings. Do you have any thoughts as to how customers might be able to set these values conditionally for selective array items within the YAML?

charts/rstudio-library/files/helm-2/service.tpl
colearendt: Yeah, we can do that if we want. I'm open to thoughts - I was just counting on the data having the default in it 🤷 We are in control of the defaults in the `values.yaml`, which would be a very natural place to see "what is the default" without having to go read these templates.

Plus it keeps things a smidge simpler in the template
https://github.com/rstudio/helm/blob/2bfb94d8b3331d10a3a78a74040a6b4433a6be4c/charts/rstudio-workbench/values.yaml#L399-L404

sagerb: Totally agree on keeping it simpler. As long as you can control the defaults, I'd lean the direction you had proposed!

sagerb:
> This also is using the templates to set global settings. Do you have any thoughts as to how customers might be able to set these values conditionally for selective array items within the YAML?

@colearendt Didn't want you to miss this question.

colearendt:
Oops @sagerb I definitely did miss that question, sorry 😅 

So you're asking how someone would choose certain jobs to have certain values set? I.e. imagePullPolicy: Always for some items, not for others?

I definitely care a lot about that case, but I think the line gets more fuzzy / advanced there. That said, I am suspicious that (if we handle our whitespace properly) we could do something like this:
```
sessionTemplate:
  imagePullPolicy: |
    {{- if eq .Job.appId 1234 }}
      Always
    {{- else }}
      IfNotPresent
    {{- end }}
```

To be clear, this is a "nice to have" IMO and perhaps something that we don't target as a first-class-citizen until we are really happy with how things are laid out

colearendt:
That said, I am pretty happy with how this is shaping up for RSC. We could still decide to hide these defaults more by generating internally (i.e. which would make determining the init image version from the chart version easier), but it feels natural-ish if we can get the right ability to generate feedback quickly:

https://github.com/rstudio/helm/pull/216/files#diff-e093ec567402374e44d1c256de177aa70486f565cb50ef6927b52e804b9d19b3

PR #214, opened 2022-05-31
Add sharedStorage.mountContent and deprecation warning

dbkegley:


dbkegley:
Up to you, I could go either way. I opted not to check for it on the first pass because the validation felt a little bit awkward since `sharedStorage.mountContent` _might_ derive `config.Launcher.DataDirPVCName`:

```
{{- if and (eq .Values.sharedStorage.mountContent true) (.Values.config.Launcher.DataDir) }} -> error
or
{{- if and (.Values.config.Launcher.DataDirPVCName)  (.Values.config.Launcher.DataDir) }} -> error
```

colearendt (approved):
Ahh yeah, that is less than ideal. I'm with you - let's punt on it for now and see if it becomes an issue - the Connect warning / error should be a reasonably fast cycle-time if someone does forget

Issue #223, opened 2022-06-23
nodeSelector and affinity consistency


colearendt:
The charts are currently all over the place with consistency on affinity and nodeSelectors

Related to #17 and #192

rodrigobrim:
The values.yaml/docs does not have any reference to the [selectorLabels](https://github.com/rstudio/helm/blob/b878e60519b56d357d421041f87de214faab0da6/charts/rstudio-pm/templates/deployment.yaml#L18) config.

dbkegley:
Another thing to note: tolerations are also handled differently. RSW appears to support a top-level `tolerations` field, while the Connect chart does not yet support tolerations. It would be great to consolidate these into a single approach or use a library helper for some of these common configurations files on the top-level `pod`

pat-s:
Unfortunately we're still facing placement issues on our mixed arch cluster: we have set 

```yml
config:
  profiles:
    launcher.kubernetes.profiles.conf:
       "*":
         placement-constraints: "kubernetes.io/arch=amd64"
```

and we see 

![image](https://user-images.githubusercontent.com/8430564/222502784-420cdb79-66c8-4a97-9fb8-6105f529b501.png)

yet sessions still get scheduled on the arm64 nodes from time to time.

I double-checked that the entry is reflected correctly into `/mnt/configmap/rstudio/launcher.kubernetes.profiles.conf` in the WB server pod:

```
[*]
placement-constraints=kubernetes.io/arch=amd64
```

Is there anything I am still misconfiguring here? 🤔

colearendt:
I think the way you want to set these are:
```
config:
  profiles:
    launcher.kubernetes.profiles.conf:
       "*":
         placement-constraints: 
           - "kubernetes.io/arch:amd64"
```

But it will only allow selection of amd64 - default will still be to not specify a placement constraint 😞

pat-s:
Thanks! Did so but that did not cause a change in `/mnt/configmap/rstudio/launcher.kubernetes.profiles.conf` (even after recreating the pod), i.e. the config specification is still looking like in my previous post. 

Let's see how it unfolds as it's hard to test directly as most sessions first go to an amd64 node anyway if there's space (due to how the cluster is set up WRT to node resources).

colearendt:
Right 😭 This is pretty unfortunate. There is a pretty ugly workaround... which is "invert the logic in the templates" 😱 

We have some work internally to make the default more configurable. 

[...]

PR #225, opened 2022-06-24
add rstudio-pm configuration values

### LINKED ISSUE #206: rstudio-pm pod metadata labels
Could you add a way to add metadata label to the pod template using the rstudio-pm helm chart values.yaml, similar to workbench and connect charts?

apiVersion: v1
kind: Pod
metadata:
  name: demo
  labels:
    aadpodidbinding: <aad_pod_identity_name>

https://github.com/rstudio/helm/blob/03b3cb610ee38a0dbdd1236cfad32dd6760ce13d/charts/rstudio-pm/values.yaml#L92

colearendt:
add rstudio-pm configuration values

also bump version and news

close #206 

Worth noting that this is related to #17 and #223 . However, `rstudio-workbench` has these values at the global level (i.e. `tolerations`, `affinity`, etc.) instead of under` pod.` We opted for `pod.` here because it seems like the better convention and this is what Connect is usingl.

### DIFF
--- a/charts/rstudio-pm/NEWS.md
+++ b/charts/rstudio-pm/NEWS.md
@@ -1,3 +1,7 @@
+# 0.3.11
+
+- Add configuration values for the pod's `labels`, `affinity`, `nodeSelector`, `tolerations`, and `priorityClassName` ([#206](https://github.com/rstudio/helm/issues/206)).
+
 # 0.3.10
 
 - The Package Manager container no longer runs as privileged by default.
--- a/charts/rstudio-pm/values.yaml
+++ b/charts/rstudio-pm/values.yaml
@@ -100,6 +100,8 @@ pod:
   serviceAccountName: false
   # -- annotations is a map of keys / values that will be added as annotations to the pods
   annotations: {}
+  # -- Additional labels to add to the rstudio-pm pods
+  labels: {}
   # -- Container [lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/)
   lifecycle: {}
   # -- the [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for the pod
@@ -203,3 +205,11 @@ config:
 
 # -- sidecar container list
 extraContainers: []
+# -- A map used verbatim as the pod's "affinity" definition
+affinity: {}
+# -- A map used verbatim as the pod's "nodeSelector" definition
+nodeSelector: {}
+# -- An array used verbatim as the pod's "tolerations" definition
+tolerations: []
+# -- The pod's priorityClassName
+priorityClassName: null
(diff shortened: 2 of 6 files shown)


charts/rstudio-pm/NEWS.md
```diff
@@ -1,3 +1,9 @@
+# 0.3.11
+
+- Add Package Manager configuration values
+  - `labels`, `sidecar`, `affinity`, `nodeSelector`, `tolerations`, `priorityClassName`
+  - close [#206](https://github.com/rstudio/helm/issues/206)
```
atheriel, line 5: Nit: I think this could all be one line.

```suggestion
- Add configuration values for the pod's `labels`, `sidecar`, `affinity`, `nodeSelector`, `tolerations`, and `priorityClassName` ([#206](https://github.com/rstudio/helm/issues/206)).
```

charts/rstudio-pm/values.yaml
```diff
@@ -100,6 +100,18 @@ pod:
   serviceAccountName: false
   # -- annotations is a map of keys / values that will be added as annotations to the pods
   annotations: {}
+  # -- Additional labels to add to the rstudio-pm pods
+  labels: {}
+  # -- An array of containers that will be run alongside the main pod
+  sidecar: false
```
atheriel, line 106: I'm pretty sure the empty array `[]` is falsy in Go templates, so can we use that instead of the (confusing) `false` here?

atheriel, line 106: What is the purpose of specifying "sidecar" containers when we already support `extraContainers`?

charts/rstudio-pm/NEWS.md
```diff
@@ -1,3 +1,7 @@
+# 0.3.11
+
+- Add configuration values for the pod's `labels`, `sidecar`, `affinity`, `nodeSelector`, `tolerations`, and `priorityClassName` ([#206](https://github.com/rstudio/helm/issues/206)).
```
atheriel, line 3: Nit: we list `sidecar` here but no longer have a key for it after the rebase.

colearendt: 🙈

atheriel:
OK, my overall view from looking around at other popular charts -- e.g. [Ingress NGINX](https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml), [Grafana](https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml), [Kube Prometheus Stack](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml), and [Cert Manager](https://artifacthub.io/packages/helm/cert-manager/cert-manager?modal=values&path=podLabels) -- is that top-level keys are the Helm community expectation for `affinity`, `tolerations`, and `nodeSelector`, so we should use that instead of nesting them under `pod`. Labels and priority class are more ambiguous so I think keeping them under `pod` is fine but I'd prefer to have `priorityClassName` as a top-level key instead, too.

colearendt:
That makes sense @atheriel ! Thanks for reviewing those!! I agree - consistency w/ the broader community is helpful. Updated

atheriel (approved):

PR #227, opened 2022-06-27
Enable the bundled R version in RSPM by default

atheriel:
The official image already ships with R to enable Git builds, but they are not turned on by default (because the configuration must point to the install location). This PR adds the bundled R version to the default configuration so that Git builds work out of the box.


colearendt (commented):
LGTM! Maybe bump the version after we merge #225 (presuming it is good to go) 😄

colearendt:
Oh yeah, for sure, just inherited from Connect (which uses the same mechanism) 😄

PR #229, opened 2022-06-27
Run Package Manager as non-root by default

atheriel:
This is a long-wanted change that brings Package Manager's behaviour on Kubernetes more in line with its non-containerised deployments.

To make this change work, this PR does the following:

* RSPM now runs as non-root and the container security context has been updated to reflect the permissions required to do so.

* There is a new top-level `enableSandboxing` setting that gives users a direct way to disable sandboxing of Git builds, which reduces the Kubernetes security requirements and should allow the RSPM chart to run on any non-OpenShift cluster without modification.

* To handle the migration of existing data (which is owned by `root`), this PR includes a new Helm hook that essentially runs `chown` on the data directory every time a user runs `helm upgrade`. Unfortunately, we can't detect when we actually need to run this migration, so it currently runs unconditionally. The hook can be disabled by setting `enableMigrations=false`; in the future when we no longer expect users to have `root`-owned data, this will become the default.

Support for handling the ownership of the encryption key file is still in progress but will probably require using a preview build.


charts/rstudio-pm/templates/deployment.yaml
```diff
@@ -86,11 +98,6 @@ spec:
             readOnly: true
             mountPath: "/home/rstudio-pm/.aws"
           {{- end }}
-          {{- if .Values.rstudioPMKey }}
```
colearendt, line 89: Is there any case where a user might want to do this instead of an env var? I guess they can always do it with `pod.volumes` or `pod.volumeMounts`?

colearendt (commented):
This is really exciting! Well done!

charts/rstudio-pm/values.yaml
```diff
@@ -193,13 +202,12 @@ serviceMonitor:
 
 # -- config is a nested map of maps that generates the rstudio-pm.gcfg file
 config:
+  Server:
+    RVersion: "/opt/R/3.6.2/"
   HTTP:
     Listen: :4242
   Metrics:
     Enabled: true
-  Launcher:
```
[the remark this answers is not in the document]
atheriel: Yes: the upcoming Package Manager release includes a clear and helpful error message indicating what they need to change. See https://github.com/rstudio/package-manager/issues/8324 and Greg's [comment](https://github.com/rstudio/package-manager/issues/8324#issuecomment-1168046664).

charts/rstudio-pm/templates/deployment.yaml
atheriel: No, if you are not running as root it's basically impossible to use it as a volume mount, and I can't think of a scenario where you would prefer a volume mount to an environment variable.

charts/rstudio-pm/values.yaml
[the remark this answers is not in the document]
colearendt: Something we have done in the helm chart for Connect to give people a very clear UX is to alert on some of these things in the helm chart. Does this seem like one that might be worthwhile? 

https://github.com/rstudio/helm/blob/72f5c8344a9cc9a3a8b080e4951b51b54e4e6195/charts/rstudio-connect/templates/NOTES.txt#L41

charts/rstudio-pm/values.yaml
```diff
@@ -29,7 +29,7 @@ sharedStorage:
 
 image:
   # -- the repository to use for the main pod image
-  repository: rstudio/rstudio-package-manager
+  repository: rstudio/rstudio-package-manager-preview
```
colearendt, line 32: TODO: need to switch this back

charts/rstudio-pm/templates/post-upgrade-chown-job.yaml
```diff
@@ -0,0 +1,50 @@
...
+          capabilities:
+            add: ["CHOWN", "DAC_OVERRIDE"]
+            drop: ["ALL"]
+        volumeMounts:
+          - name: rstudio-pm-data
+            mountPath: /var/lib/rstudio-pm
+      volumes:
+      - name: rstudio-pm-data
+        persistentVolumeClaim:
+          claimName: {{default (print (include "rstudio-pm.fullname" .) "-shared-storage" ) .Values.sharedStorage.name }}
```
colearendt, line 49: We can probably dispense with running this if there is no PVC? What if they are specifying their own `volume` manually?

charts/rstudio-pm/templates/post-upgrade-chown-job.yaml
```diff
@@ -0,0 +1,50 @@
...
+# any persistent storage.
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "rstudio-pm.fullname" . }}-chown-shared-storage
+  namespace: {{ $.Release.Namespace }}
+  labels:
+    {{- include "rstudio-pm.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-upgrade
```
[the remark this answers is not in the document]
colearendt: TODO: we discussed this and decided to simplify the rollOver policy

charts/rstudio-pm/templates/post-upgrade-chown-job.yaml
colearendt: There is already a check at the top for this 😅

charts/rstudio-pm/templates/post-upgrade-chown-job.yaml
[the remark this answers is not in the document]
colearendt: Actually, we decided to remedy this problem with docs 😄 The risk is somewhat minimal, the workaround to cache the version with `lookup` feels fragile, and the remedy fairly straightforward.

charts/rstudio-pm/values.yaml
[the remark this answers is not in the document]
colearendt: Implemented such a check

charts/rstudio-pm/values.yaml
```diff
@@ -110,12 +121,15 @@ pod:
   # -- the [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for the pod
   securityContext: {}
   # -- the [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for the main Package Manager container
+  #    Evaluated as a template.
```
atheriel, line 124: Nit: the fact that this is on another line creates a weird whitespace break in the generated `README`.

charts/rstudio-pm/templates/NOTES.txt
```diff
@@ -11,6 +11,17 @@ NOTE: `.Values.config.Licensing.LicenseType` is configured manually. Normally, w
 Please consider removing this configuration value.
 {{- end }}
 {{- end }}
+
+{{- if hasKey .Values.config "Launcher" }}
+  {{- if hasKey .Values.config.Launcher "ServerUser" }}
+    {{- if eq (get .Values.config.Launcher "ServerUser") "root" }}
+      {{- if .Values.rootCheckIsFatal }}
+        {{- fail "\n\n`config.Launcher.ServerUser=root`, but the public helm chart and docker image now run Package Manager as non-root. Please remove the `root` values from your values file.\n\nDANGER: set `.Values.rootCheckIsFatal` to bypass this error." }}
```
atheriel, line 19: I'm not sure we need to fail here, I think a warning is fine. Package Manager itself has been updated to log a useful error message in this case that should explain the remedy.

colearendt: I am inclined to just remove this whole block - it's easier that way for sure. The one reason I have pause is because anyone who copy / pasted the values file (I have seen this in practice a good bit) to set their defaults _will_ hit this issue, so the faster turnaround cycle seems preferable. i.e. we set them up for this problem w/ not-great values before.

I'm not sure why this is a convention that folks adhere to... but it makes me wonder whether we should keep the extra check or not.

tylfin (approved):

colearendt (approved):
I'm going to go ahead and merge this - we can fix up things on the formal 0.4.0 release if needed

PR #231, opened 2022-06-28
Add Service Account support and related doc improvements for Package Manager

atheriel:
This PR adds `ServiceAccount` creation for the Package Manager pod, primarily to support the use of [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) for S3 access. Not only is this the best-regarded choice from a security perspective, it is also what we use internally.

Relatedly, I've also totally overhauled the S3-related section of the README to guide users towards that path.

cc @rodrigobrim

### DIFF
--- a/charts/rstudio-pm/README.md
+++ b/charts/rstudio-pm/README.md
@@ -46,22 +46,44 @@ This chart requires the following in order to function:
 
 ## S3 Configuration
 
-Package Manager can be configured to store data in S3 buckets (see the [Admin Guide](https://docs.rstudio.com/rspm/admin/files-directories/#data-destinations)
-for more details). When configured this way, AWS access credentials are required. You must set the `awsAccessKeyId` and `awsSecretAccessKey` chart values
-to ensure that RSPM will be able to authenticate with your configured S3 buckets.
+Package Manager [can be configured to store its data in S3
+buckets](https://docs.rstudio.com/rspm/admin/files-directories/#data-destinations),
+which eliminates the need to provision shared storage for multiple replicas. A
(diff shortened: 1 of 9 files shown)


charts/rstudio-pm/README.md
```diff
@@ -119,7 +141,7 @@ The Helm `config` values are converted into the `rstudio-pm.gcfg` service config
 | pod.labels | object | `{}` | Additional labels to add to the rstudio-pm pods |
 | pod.lifecycle | object | `{}` | Container [lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) |
 | pod.securityContext | object | `{}` | the [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for the pod |
-| pod.serviceAccountName | bool | `false` | serviceAccountName is a string representing the service account of the pod spec |
+| pod.serviceAccountName | string | `""` | Deprecated, use `serviceAccount.name` instead |
```
colearendt, line 144: Should we warn about this if it is used, to make the deprecation a bit more formal? If we throw it in `NOTES.txt`, that could be a good place to establish this as a convention for how to migrate.

colearendt (commented):
This looks great! We should also add a `NEWS.md` entry

charts/rstudio-pm/README.md
atheriel: Good idea.

rodrigobrim (approved):

PR #232, opened 2022-06-29
Add the hability to remove the "helm.sh/resource-policy: keep" default annotation from the pvc

rodrigobrim:
### Problem

Today, the PVC created by the rstudio-pm chart is protected (`helm.sh/resource-policy: keep` annotation). However, in our deployment, we need to delete this resource to free the underlying PV, which our provisioning software controls.

### Solution

Add the `pvcPolicy: keep` to the values.yaml by default, allowing the possibility to comment it out.


atheriel:
I think this brings up a good point. We use dynamically-provisioned PVCs here, and it is a known limitation of Kubernetes that [dynamically-provisioned PVCs don't support setting the `ReclaimPolicy`](https://github.com/kubernetes/kubernetes/issues/30606).

By setting this annotation unconditionally, we are assuming that users will want to keep their data around, which is probably not always true. However, the common community answers -- "make sure you patch the associated PV to set the `ReclaimPolicy` before uninstalling the chart" or "create your own `StorageClass` and use that instead of the default -- are pretty awkward, too. It's not crazy to me that we should try to prevent accidental data loss by default.

An alternative to this proposal is to expose a `sharedStorage.annotations` setting that includes this policy annotation by default. That might be a good option if there are other scenarios where users would want to add annotations to their persistent volume claims -- for example, some drivers [support encrypting volumes this way](https://docs.portworx.com/key-management/kubernetes-secrets/pvc-encryption-using-annotations/). From a [quick search of Helm's old chart repository](https://github.com/helm/charts/search?q=%22helm.sh%2Fresource-policy%22%3A+keep), exposing the `annotations` for a PVC also seems like the more common choice, though both are present.

rodrigobrim:
Good points, Cole. Can I assume the annotation solution instead of the proposed first?

Can I apply it to all charts?

colearendt:
I like the annotation solution as well!

rodrigobrim:
Hey @atheriel, could you take a look at the last change?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
tylfin, line 1: We probably want to add this directory to the `.gitignore` file

rodrigobrim: Done.

charts/rstudio-connect/NEWS.md
```diff
@@ -1,3 +1,7 @@
+# 0.2.35
+
+- Add the ability to set annotations to the PVC.
```
atheriel, line 3: Here and elsewhere: avoid using acronyms like "PVC" in user-facing docs. We should use some variant of "Persistent Volume Claim" instead.

charts/rstudio-connect/values.yaml
```diff
@@ -25,6 +25,9 @@ sharedStorage:
   requests:
     # -- The volume of storage to request for this persistent volume claim
     storage: "10Gi"
+  # -- Define the annotations for the PVC resource
```
atheriel, line 28: Suggested docstring, here and elsewhere:

```suggestion
  # -- Annotations for the PersistentVolumeClaim
```

.gitignore
```diff
@@ -4,3 +4,4 @@ charts/**/charts/
 /.cr-release-packages
 bin/**
 !bin/README.md
+.vscode/
```
atheriel, line 7: Nit: missing line break.

atheriel:
Nit: if you are going to include the `.gitignore` commit, please rebase the rest of the history into a more coherent commit so it will be understandable when merged.

colearendt:
Ultimately it is _best_ to leave editor-specific ignore statements in `~/.config/git/ignore` (by default) instead of in project specific directories. https://git-scm.com/docs/gitignore (since everyone uses different tools / editors)

charts/rstudio-connect/templates/pvc.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   name: {{default (print (include "rstudio-connect.fullname" .) "-shared-storage" ) .Values.sharedStorage.name }}
   namespace: {{ $.Release.Namespace }}
   annotations:
-    "helm.sh/resource-policy": keep
+{{ .Values.sharedStorage.annotations | toYaml | indent 4}}
```
colearendt, line 8:
```suggestion
    {{- .Values.sharedStorage.annotations | toYaml | nindent 4}}
```
I personally prefer this pattern because it makes indentation look proper in the template and in the output (same elsewhere)

charts/rstudio-connect/NEWS.md
tylfin:
```suggestion
- Add the ability to set annotations to the Persistent Volume Claim.
```

charts/rstudio-pm/values.yaml
```diff
@@ -23,6 +23,9 @@ sharedStorage:
   requests:
     # -- the volume of storage to request for this persistent volume claim
     storage: "10Gi"
+  # -- Define the annotations for the PVC resource
```
tylfin, line 26:
```suggestion
  # -- Define the annotations for the Persistent Volume Claim resource
```

charts/rstudio-workbench/NEWS.md
```diff
@@ -1,3 +1,7 @@
+# 0.5.16
+
+- Add the ability to set annotations to the PVC.
```
tylfin, line 3:
```suggestion
- Add the ability to set annotations to the Persistent Volume Claim.
```

charts/rstudio-workbench/values.yaml
```diff
@@ -43,6 +43,9 @@ sharedStorage:
     storage: "10Gi"
   # -- selector for PVC definition
   selector: {}
+  # -- Define the annotations for the PVC resource
```
tylfin, line 46:
```suggestion
  # -- Define the annotations for the Persistent Volume Claim resource
```

charts/rstudio-connect/values.yaml
```diff
@@ -25,6 +25,9 @@ sharedStorage:
   requests:
     # -- The volume of storage to request for this persistent volume claim
     storage: "10Gi"
+  # -- Annotations for the PersistentVolumeClaim
```
tylfin, line 28:
```suggestion
  # -- Annotations for the Persistent Volume Claim
```

charts/rstudio-pm/NEWS.md
```diff
@@ -1,3 +1,7 @@
+# 0.3.12
+
+- Add the ability to set annotations to the PVC.
```
tylfin, line 3:
```suggestion
- Add the ability to set annotations to the Persistent Volume Claim.
```

atheriel (approved):

PR #242, opened 2022-07-11
DRAFT: add an adhoc chart-releaser

colearendt:
add an adhoc chart-releaser that is triggered by the `/release-adhoc`…

This allows for releasing (in an adhoc fashion) from a branch. Unfortunately the behavior and how this will work for non-devel releases is not super well-defined. If chart-releaser adds support for filtering to only certain releases, that would be helpful to make this safer: https://github.com/helm/chart-releaser/issues/196

In particular, it is most concerning if chart A moves to `--devel` version on the `dev` branch, chart B moves to a newer version on `main`, and then CI runs based on a `/release-adhoc` comment. It would also be concerning if chart A or chart B move to a newer version on a branch and then publish a "non-devel" version from the branch.

It requires being an `OWNER` on the repository, although it is unclear if this correlates to `ADMIN` privileges within the organization.

### DIFF
--- a/.github/workflows/chart-releaser-adhoc.yaml
+++ b/.github/workflows/chart-releaser-adhoc.yaml
@@ -1,0 +1,47 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+name: Release Charts (adhoc)
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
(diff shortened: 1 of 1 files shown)


bdeitte:
Given the cavaets already raised in here and how old this is, closing this out.

PR #249, opened 2022-07-21
bump connect version to 2022.07.0

colearendt:
Also remove the `-preview` suffix for image repositories, since 2022.07 (preview builds) had already been running on main in a `--devel` context

### DIFF
--- a/charts/rstudio-connect/Chart.yaml
+++ b/charts/rstudio-connect/Chart.yaml
@@ -1,8 +1,8 @@
 name: rstudio-connect
 description: Official Helm chart for RStudio Connect
-version: 0.2.39-alpha03
+version: 0.2.39-alpha04
 apiVersion: v2
-appVersion: 2022.07.0-dev-428
+appVersion: 2022.07.0
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
 home: https://www.rstudio.com
 sources:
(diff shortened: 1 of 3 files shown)


charts/rstudio-connect/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 name: rstudio-connect
 description: Official Helm chart for RStudio Connect
-version: 0.2.39-alpha03
```
whimsicallyson, line 3: based on previous helm chart examples, I would have assumed this went to `0.2.40`. How do we know what to change it to?
I'd like to add more detail to the wiki page to better document!

charts/rstudio-connect/Chart.yaml
```diff
@@ -1,8 +1,8 @@
...
+version: 0.2.39-alpha04
 apiVersion: v2
-appVersion: 2022.07.0-dev-428
```
whimsicallyson, line 5: similarly, why is there a `-dev-428` suffix on here? is that normal, or would we usually have something like `2022.06.2` since that was the last released version?

charts/rstudio-connect/Chart.yaml
colearendt: Yeah, normally 0.2.40 would have been the right call 😄 This release is special because it is the launching of the trebuchet beta. You can leave the docs at suggesting 0.2.40 😄 

(Basically we "already" landed 2022.07 in a context that is "hidden." This leaves it hidden until we finish testing a bunch of chart changes on Colorado)

charts/rstudio-connect/Chart.yaml
colearendt: Same explanation as above 😄 Yes - normally you would have seen 2022.06.2 getting removed

Issue #275, opened 2022-09-07
Issues mounting PVC via `job-json-overrides` into session pod


pat-s:
I am struggling to mount a custom PVC into the session image via `job-json-overrides`.

I followed the example in https://github.com/rstudio/helm/blob/main/docs/customize.md (the last which shows the PVC) but launching a new sessions fails with

```
│ 2022-09-07T13:18:46.969189Z [rstudio-kubernetes-launcher, log-source: Kubernetes] ERROR system error 71 (Protocol error) [description: Invalid status returned from  │
│ create resource at /apis/batch/v1/namespaces/rsw-efs/jobs: Unprocessable Entity - {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Job │
│ .batch \"session-e10d7cfc78a3127bfedcf-patrick---rstudio-session-vz7w4\" is invalid: [spec.template.spec.volumes[3].persistentVolumeClaim.claimName: Required value, │
│  spec.template.spec.containers[0].volumeMounts[3].name: Not found: \"rsw-efs-renv-cache-volume\"]","reason":"Invalid","details":{"name":"session-e10d7cfc78a3127bfed │
│ cf-patrick---rstudio-session-vz7w4","group":"batch","kind":"Job","causes":[{"reason":"FieldValueRequired","message":"Required value","field":"spec.template.spec.vol │
│ umes[3].persistentVolumeClaim.claimName"},{"reason":"FieldValueNotFound","message":"Not found: \"rsw-efs-renv-cache-volume\"","field":"spec.template.spec.containers │
│ [0].volumeMounts[3].name"}]},"code":422}  
```

I am suspecting namespace issues but also tried to put the PVC into the default namespace without success.

Here's my yaml definition - am I doing something wrong?

```yml
config:
  profiles:
    launcher.kubernetes.profiles.conf:
        job-json-overrides:
          - name: rsw-efs-renv-cache-volume
            target: /spec/template/spec/volumes/-
            json:
              name: rsw-efs-renv-cache-volume
              # namespace: rsw-efs # does not help
              persistentVolumeClaim:
                name: rsw-efs-renv-cache
                # namespace: rsw-efs # does not help
          - name: rsw-efs-renv-cache-volume-mount
            target: /spec/template/spec/containers/0/volumeMounts/-
            json:
              name: rsw-efs-renv-cache-volume
              mountPath: /renv
```

[...]

PR #276, opened 2022-09-12
Update `name` specification for PVC mounts

pat-s:
see https://github.com/rstudio/helm/issues/275#issuecomment-1241332262


colearendt (approved):
Great catch! Thank you!

PR #213, opened 2022-05-27
DRAFT: Use data templates with rsw

colearendt:

PR #278, opened 2022-09-14
fix(rstudio-workbench): don't change ownership of projected volumes

jgoldschrafe:
The Workbench launcher's startup script currently attempts to assume ownership of all files underneath `/var/lib/rstudio`, which fails when the contents of those directories are read-only projected volumes (i.e. mounted from ConfigMap or Secret resources).

This change backports the prestart-workbench.bash fix from #213 in order to make it available immediately to users who are generating and volume-mounting Job and Service templates, while the larger PR is polished and wrapped up.

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-workbench
 description: Official Helm chart for RStudio Workbench
-version: 0.5.19
+version: 0.5.20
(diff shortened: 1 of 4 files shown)


colearendt (approved):
Thanks for submitting this! I added a version bump and news entry to keep CI happy 😄

colearendt:
Also batched in another change (since bumping the version on each PR can be tedious) in accordance with #279

PR #294, opened 2022-11-07
Explicitly setting service account to empty string

meatballhat:
which *must be done* according to an amazing thread of issues reported by humans bitten by this intentional inconsistency in kubernetes "apply" behavior :tada:

As referenced in the body of this change, one may trace back through this gotcha from this issue comment:

https://github.com/kubernetes/kubernetes/issues/108208#issuecomment-1262269204

and also see this mention in the kubernetes docs in a "Note" callout near the end of this section:

https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-multiple-service-accounts


colearendt:
Merged in #292

PR #301, opened 2022-11-16
Fix logging config

gbrlsnchs:


aronatkins (commented):
This change will need to be included with the 2022.11.0 version bump and will need to also adjust the helm chart version and include NEWS about the update.

Lytol:
Closed in favor of #302

Issue #293, opened 2022-11-05
Add securityContext to templating


colearendt:
This is an important miss 😞

pat-s:
I think I am right now stumbling over this.

On AKS with a policy denying priviliged execution, I am struggling to set the `securityContext` of the `packrat-restore` job.batch to `null`. By default it is set to 

```yml
securityContext:                                                          
  runAsGroup: 999                                                         
  runAsUser: 999
```

which results in a denied execution by the gatekeeper.

Setting

```yml
launcher:
  useTemplates: true
  enabled: true
  templateValues:
    job:
      securityContext: null
```

doesn't change yet.

pat-s:
AFAICS the job definition of `securityContext` is currently not accepting templateData input:

https://github.com/rstudio/helm/blob/9d58e030b0eafac5a4834a131685d62dd1c9d0c5/charts/rstudio-connect/files/job.tpl#L111-L115

colearendt:
Howdy @pat-s! Thanks for the practical example here! Can I ask what error message the gatekeeper is throwing? I am surprised that it is upset about `runAsUser` and `runAsGroup` - that is an important piece of defining an unprivileged job in my understanding 😮  (it is basically saying: make sure you don't run this as root. Instead use uid/gid 999/999)

pat-s:
Apologies for making rookie mistakes here, this is the first time I am doing such a deep dive on the non-root executions 🙃 

The gatekeeper throws

```
{"level":"info","ts":1669115646.2458007,"logger":"webhook","msg":"denied admission","process":"admission","event_type":"violation","constraint_name":"azurepolicy-k8 │
│ sazurev3noprivilegeescalatio-11f7b53155b7de191e38","constraint_group":"constraints.gatekeeper.sh","constraint_api_version":"v1beta1","constraint_kind":"K8sAzureV3No │
│ PrivilegeEscalation","constraint_action":"dryrun","resource_group":"","resource_api_version":"v1","resource_kind":"Pod","resource_namespace":"connect","resource_nam │
│ e":"run-shiny-application-vcr6c-jjz7q","request_username":"system:serviceaccount:kube-system:job-controller"}
```

The `job.batch` itself is running with

```yml
securityContext:                                                          
  runAsGroup: 999                                                         
  runAsUser: 999  
```

[...]

PR #307, opened 2022-11-22
Connect: Allow overriding `securityContext` for jobs

pat-s:
relates #293 

Disclaimer: untested, just a first take on allowing to set `securityContext` for jobs from template data. I don't know if the presented notation will work in practice.

I essentially just added an if clause which sets the `securityContext` if it's defined and otherwise uses the previous definition for the dict.

Please modify and correct as needed :)

### DIFF
--- a/charts/rstudio-connect/files/job.tpl
+++ b/charts/rstudio-connect/files/job.tpl
@@ -94,23 +94,27 @@ spec:
         {{- end }}
       {{- end }}
       {{- $securityContext := dict }}
-      {{- if .Job.container.runAsUserId }}
-        {{- $_ := set $securityContext "runAsUser" .Job.container.runAsUserId }}
-      {{- end }}
-      {{- if .Job.container.runAsGroupId }}
-        {{- $_ := set $securityContext "runAsGroup" .Job.container.runAsGroupId }}
-      {{- end }}
-      {{- if .Job.container.supplementalGroupIds }}
-        {{- $groupIds := list }}
-        {{- range .Job.container.supplementalGroupIds }}
-          {{- $groupIds = append $groupIds . }}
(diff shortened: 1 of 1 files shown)


colearendt:
Thanks so much for tagging in here! Unfortunately this will be a bit tricky to implement since all of those pieces you removed are used by the Launcher API itself (i.e. that is the mechanism that Connect / Workbench use to set the RunAsUser, RunAsGroup, etc.). So we probably can't change the template in this way since it would globally disable functionality within the products 🙈 

That said, you can definitely make modifications of this sort to your environment if that is helpful! I keep intending to create a way to more easily override templates if you wanted to patch in your own. In a particular environment, it is totally reasonable to disable functionality that is not needed / not in use / causing problems / etc. 😄 

I asked a question on #293 to see if I can get a bit more understanding about what is failing for you and what would be required to fix!

colearendt:
Woops - apologies for the delayed reply here!

Admittedly until #309 it was a bit hard to make changes to the templates without the chart for Connect. 

What you can do as of #309 in any particular Connect deployment is use:

```
launcher:
  useTemplateValues: true
  useDefaultTemplates: false
  extraTemplates:
    job.tpl: |
      your-own-inline-template
    service.tpl: |
      your-own-inline-template
```

This will "undo" our in-chart template and allow you to specify your own. Admittedly, templates are a bit complex, so it's probably ideal to use our templates as a baseline / input to this process. The tricky part of template customization usually becomes "revisiting the process" or "handling when we update the templates" 😞 `vimdiff` has been an invaluable tool in this effort for my own work, and we are definitely keeping an eye out for ways to simplify this process 😄

colearendt:
Closing in favor of #309

PR #309, opened 2022-11-24
Update templates

### LINKED ISSUE #251: Fix bug in serviceAccountAnnotations
In the `rstudio-workbench` chart, we document `rbac.serviceAccount.annotations` but use `serviceAccountAnnotations`: https://github.com/rstudio/helm/blob/730e06c19a47ff711f575c8a4c2e055a02abbd3a/charts/rstudio-workbench/templates/rbac.yaml#L7

We should resolve this discrepancy and check the other charts 🙈

### LINKED ISSUE #271: RSW: Tolerations and affinity for session pods
Azure SPOT instances require setting certain affinity and tolerations for a pod to be able to scheduled on it. See https://docs.microsoft.com/en-us/azure/aks/spot-node-pool#verify-the-spot-node-pool.

Here's the required config:

```yml
  tolerations:
  - key: "kubernetes.azure.com/scalesetpriority"
    operator: "Equal"
    value: "spot"
    effect: "NoSchedule"
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: "kubernetes.azure.com/scalesetpriority"
            operator: In
            values:
            - "spot"
```

AFAICS one can currently only define these for the server pods (which then get scheduled fine). Session pods error with

```
Warning  FailedScheduling  5m33s  default-scheduler  0/2 nodes are available: 1 Insufficient memory, 1 node(s) had untolerated taint {kubernetes.azure.com/scalesetpriority: spot}.
```

### LINKED ISSUE #283: RSC: affinity for launcher pods
Pods deployed via the launcher (using off-host execution) do not respect the `pod.affinity` definition in the helm chart. Would be great if the setting could be honored or if launcher pods could even have a distinct affinity definition.

Related to #223

### LINKED ISSUE #293: Add securityContext to templating

colearendt:
update the templates and associated charts

- add securityContext (close #293 , supersede #307 )
- add affinity and tolerations (close #271 and close #283 )
- fix the bug in `serviceAccountName` (close #251 )

Also... reviewing this kind of thing is really tricky 😬 Especially since all of our attempted snapshots are mostly just random version number changes, and we always forget to update them because they are not in CI 😭

### DIFF
--- a/charts/rstudio-connect/Chart.yaml
+++ b/charts/rstudio-connect/Chart.yaml
@@ -1,8 +1,8 @@
 name: rstudio-connect
 description: Official Helm chart for RStudio Connect
-version: 0.3.7
+version: 0.3.8
 apiVersion: v2
-appVersion: bionic-2022.11.0
+appVersion: 2022.11.0
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
 home: https://www.rstudio.com
 sources:
--- a/charts/rstudio-connect/NEWS.md
+++ b/charts/rstudio-connect/NEWS.md
@@ -6,6 +6,7 @@
   - Fix typo in `serviceAccountName` implementation ([#251](https://github.com/rstudio/helm/issues/251))
   - Add `affinity` and `tolerations` ([#271](https://github.com/rstudio/helm/issues/271) and [#283](https://github.com/rstudio/helm/issues/283))
 - Add an `image.tagPrefix` value to make customizing the operating system easier
+- Add a `launcher.defaultInitContainer.tagPrefix` value to make customizing the operating system easier
 
 # 0.3.7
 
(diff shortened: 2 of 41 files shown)


colearendt:
Just refactored to use `v2` templates for Workbench - it cannot handle the `2.1.0` version yet 😞 (needs a new launcher)

colearendt:
A follow-up note for anyone who happens upon this PR later... it's kinda ugly. The main reasons are:
- snapshot testing (that doesn't ignore version numbers) is messy
- the templates are not "diff'd" because we keep the history, so it's hard to see changes
- the templates are copied into three different places (four in this case, really)
- we coupled a handful of changes together. We could have done this with a long-lived feature branch and PRs into it... instead we just smooshed it all in here 😄

PR #298, opened 2022-11-10
Edit newlines across a few files

AlexMapley:
## Background
I copied these charts over to https://github.com/rstudio/platform-labs, to help prepare for a possible Mercury Chart configuration https://github.com/rstudio/platform-labs/pull/3/files.

We'd like to contribute to https://github.com/rstudio/helm now that we're using/copy+pasting the code. Hopefully these newline edits are all welcome, at least as a start.

## Summary
- Edits a few newlines, mostly removing duplicates but also adding in one case

PR #310, opened 2022-11-24
fix package manager image reference to target `bionic-`

colearendt:
Follow up to #304 

I think this is our first change since images started having a `bionic-` prefix. I was tempted to go ahead and take this opportunity to switch to `jammy-`... but with RSPM recently going through the root -> non-root switch, with the change to Posit, and Thanksgiving, I thought maybe we should be more cautious about such a change (i.e. scope it before we make the change).

NOTE: also plucked a newline from #298 to prevent another version bump


colearendt:
Going to go ahead and merge since the chart is broken at present without this image fix.

PR #315, opened 2022-12-01
Pvc options for RSC and RSPM

colearendt:
add volumeName and selector options for sharedStorage on both the Connect and RSPM charts

This is particularly useful for mapping PVCs to a particular PV


aronatkins (approved):
Do we usually update sample values files to make sure that the chart renders valid YAML with/without values?

dbkegley (approved):

colearendt:
Woops 🙈 Yes. When possible haha. I did that for the Workbench implementation and forgot it here. Thanks for catching!

Issue #280, opened 2022-09-15
Add CI to be sure that images exist


colearendt:
Related to #279 - it is possible for image naming changes, ordering issues, etc. to impact our charts if we do not verify that images exist. We should find a way to verify this within CI - perhaps by fixing integration testing (i.e. #100 )

PR #316, opened 2022-12-05
Posit Package Manager Chart does not reference an existing container at the moment

michaelmayer2:
Currently the [Posit Package Manager Chart](https://github.com/rstudio/helm/blob/main/charts/rstudio-pm/Chart.yaml#L22) does not seem to reference any container that exists on dockerhub. We reference `rstudio-package-manager:bionic-2022.11.2-18` but in fact no `-18` exists, only `--{git commit id}` tags. 

This PR is changing the reference to `rstudio-package-manager:bionic-2022.11.2`.

Additionally a typo in the NEWS section where we claim that we changed version of [Post Package Manager](https://github.com/rstudio/helm/blame/main/charts/rstudio-pm/NEWS.md#L13) is part of this PR as well.

### DIFF
--- a/charts/rstudio-pm/Chart.yaml
+++ b/charts/rstudio-pm/Chart.yaml
@@ -19,7 +19,7 @@ dependencies:
 annotations:
   artifacthub.io/images: |
     - name: rstudio-package-manager
-      image: rstudio/rstudio-package-manager:bionic-2022.11.2-18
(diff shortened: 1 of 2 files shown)


colearendt (approved):
Good catch! Highlights the need for #280 ! 

I made a couple of tweaks just to make CI happy

PR #325, opened 2022-12-15
Add launcher 2.3.0 templates

dbkegley:
- Add launcher `default/2.3.0/job.tpl` and `default/2.3.0/service.tpl`
- Add launcher `helm/2.3.0-v1/job.tpl` and `helm/2.3.0-v1/service.tpl`

### Verification

```bash
cd examples/launcher-templates

# view launcher-only modifications
vimdiff default/2.1.0/service.tpl default/2.3.0/service.tpl 
vimdiff default/2.1.0/job.tpl default/2.3.0/job.tpl 

# view helm-only modifications
vimdiff helm/2.1.0-v2/service.tpl helm/2.3.0-v1/service.tpl
vimdiff helm/2.1.0-v2/job.tpl helm/2.3.0-v1/job.tpl

# view launcher+helm modifications
vimdiff default/2.1.0/job.tpl helm/2.3.0-v1/job.tpl 
vimdiff default/2.1.0/service.tpl helm/2.3.0-v1/service.tpl 
```

Notable template changes:

- added `app.kubernetes.io/managed-by: "launcher"` in both `job.tpl` and `service.tpl`
- resource requests and limits calculations in `job.tpl`
- serviceAccountName in `job.tpl`

PR #326, opened 2022-12-15
Bump rstudio-connect launcher templates to 2.3.0-v1

dbkegley:
Depends on https://github.com/rstudio/helm/pull/325

Creating this as a placeholder. Waiting for the 2023.01.0 Connect release to bump the chart version in this PR to avoid the merge conflicts.

### TODO

Add NEWS to the rstudio-connect chart:
-  added `app.kubernetes.io/managed-by: "launcher"` in both `job.tpl` and `service.tpl`
-  resource requests and limits calculations in `job.tpl`
-  serviceAccountName in job.tpl
- `launcher.templateValues.pod.serviceAccountName` to set the default service account for content pods

### DIFF
--- a/charts/rstudio-connect/files/job.tpl
+++ b/charts/rstudio-connect/files/job.tpl
@@ -1,6 +1,6 @@
-# Version: 2.1.0
+# Version: 2.3.0
 # DO NOT MODIFY the "Version: " key
-# Helm Version: v2
+# Helm Version: v1
 {{- $templateData := include "rstudio-library.templates.data" nil | mustFromJson }}
 apiVersion: batch/v1
 kind: Job
@@ -17,6 +17,7 @@ metadata:
     {{- end }}
     {{- end }}
   labels:
+    app.kubernetes.io/managed-by: "launcher"
     {{- with .Job.metadata.job.labels }}
     {{- range $key, $val := . }}
     {{ $key }}: {{ toYaml $val | indent 4 | trimPrefix (repeat 4 " ") }}
@@ -122,6 +123,9 @@ spec:
         {{ $key }}: {{ $val }}
         {{- end }}
       {{- end }}
+      {{- if .Job.serviceAccountName }}
+      serviceAccountName: {{ .Job.serviceAccountName }}
+      {{- end }}
       {{- with $templateData.pod.imagePullSecrets }}
       imagePullSecrets: {{ toYaml . | nindent 12 }}
       {{- end }}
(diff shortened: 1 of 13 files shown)


charts/rstudio-connect/files/job.tpl
```diff
@@ -1,6 +1,6 @@
-# Version: 2.1.0
+# Version: 2.3.0
 # DO NOT MODIFY the "Version: " key
-# Helm Version: v2
+# Helm Version: v1
```
aronatkins, line 3: Why did the helm version go backwards?

examples/launcher-templates/default/2.3.0/job.tpl
```diff
@@ -0,0 +1,213 @@
+# Version: 2.3.0
```
aronatkins, line 1: Are these templates landing in a separate PR?

aronatkins (approved):

aronatkins (commented):
Do we need permissions changes to allow service account enumeration?

examples/launcher-templates/default/2.3.0/job.tpl
dbkegley: Yes, this PR is stacked on top of: https://github.com/rstudio/helm/pull/325

dbkegley:
@aronatkins Yes, I think those permissions changes should happen in the other templates PR but waiting to hear back from cole on that. https://github.com/rstudio/helm/pull/325#issuecomment-1354815122

charts/rstudio-connect/files/job.tpl
dbkegley: These are the first set of helm modifications to the launcher 2.3.0 templates. Helm starts at v1 for each new version of the launcher templates so that we can make multiple helm template changes per launcher template version.

examples/launcher-templates/default/2.3.0/job.tpl
aronatkins: GitHub isn't showing the PR as stacked; maybe change the base branch?

dbkegley: Changed my base branch. I didn't realize GH would rebase automatically after my first PR was merged!

AlexMapley (commented):
The content of these changes lgtm, I appreciate the `app.kubernetes.io/managed-by: "launcher"` label

We can sanity test these changes in our environment too if that'd be helpful @dbkegley, as of recently we're running a fork of these charts which we could easily push these changes to: https://github.com/rstudio/platform-labs/tree/main/helm/platform-bundle/charts/rstudio-connect. If only for the `charts/rstudio-connect/files/job.tpl` changes which are a little harder to eye out

dbkegley:
@AlexMapley We were able to validate these changes outside of the helm chart during QA before we merged the launcher into Connect so I think we're okay from that side, but thank you for the offer!

AlexMapley:
Ah great @dbkegley!

charts/rstudio-connect/templates/_helpers.tpl
```diff
@@ -72,6 +72,9 @@ app.kubernetes.io/instance: {{ .Release.Name }}
     {{- else }}
       {{- $_ := set $launcherSettingsDict "KubernetesUseTemplates" "false" }}
     {{- end }}
+    {{- if .Values.launcher.templateValues.pod.serviceAccountName }}
+      {{- $_ := set $launcherSettingsDict "KubernetesDefaultServiceAccount" .Values.launcher.templateValues.pod.serviceAccountName }}
```
colearendt, line 76: This is potentially redundant with the above right? Basically we set the default in Connect _and_ in the `job.tpl`, so if Connect fails to send along the SA, then it would still get applied by the templates.

colearendt (approved):
Looks awesome! Well done!

charts/rstudio-connect/templates/_helpers.tpl
dbkegley: Yes, there's definitely some awkwardness with the interaction between `Launcher.KubernetesDefaultServiceAccount` and `.Values.launcher.templateValues.pod.serviceAccountName`, and I'm not sure what the right answer is.

It's worth noting that when connect is deployed _outside_ of helm, if `Launcher.KubernetesDefaultServiceAccount` is unset, then Connect does not set a service account on the job that gets created, meaning Kubernetes should assign a default service account.

PR #328, opened 2023-01-06
add license-related lifecycle hooks to each chart

colearendt:
TODO: bump versions, test to be sure it works, etc.

### DIFF
--- a/charts/rstudio-connect/values.yaml
+++ b/charts/rstudio-connect/values.yaml
@@ -104,6 +104,11 @@ pod:
   securityContext: {}
   # -- The containerPort used by the main pod container
   port: 3939
+  # -- Definition for pod lifecycle hooks
+  lifecycle: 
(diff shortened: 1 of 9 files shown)


charts/rstudio-connect/values.yaml
```diff
@@ -104,6 +104,11 @@ pod:
   securityContext: {}
   # -- The containerPort used by the main pod container
   port: 3939
+  # -- Definition for pod lifecycle hooks
```
aronatkins, line 107: Could you compare this to the `deactivate` function in the `startup.sh` within the default Connect container?

The `deactivate` runs after the Connect process exits; when does this run?

aronatkins (approved):
OK adding the hook, but can we explain the responsibilities of the image / container versus the responsibilities of the pod (via the helm chart)?

PR #331, opened 2023-01-24
add a handful of examples for the various charts

colearendt:
We add an example for each product using sqlite (with NOT NFS), and an example of Connect using a sidecar sssd so that a dedicated Connect build is not required.

We should really test that these examples work before we add them to the repo

### DIFF
--- a/examples/connect/connect-with-sidecar-sssd.yaml
+++ b/examples/connect/connect-with-sidecar-sssd.yaml
@@ -1,0 +1,116 @@
+# WARNING: no shared storage, so no persistence for this system
+sharedStorage:
+  enabled: false
+
+pod:
+  volumes:
+    - name: sssd-emptydir
+      emptyDir: {}
+    - name: sssd-config
+      configMap:
+        name: sssd-config
+        defaultMode: 0600
+    - name: pam-config
+      configMap:
+        name: pam-config
+        defaultMode: 0755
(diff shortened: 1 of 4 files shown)


examples/connect/connect-with-sidecar-sssd.yaml
```diff
@@ -0,0 +1,116 @@
...
+  sidecar:
+    - name: sssd
+      image: "some-sssd-image"
+      imagePullPolicy: "IfNotPresent"
+      securityContext: {}
+      volumeMounts:
+        - name: sssd-emptydir
+          mountPath: /var/lib/sss
+        - name: sssd-config
+          mountPath: /etc/sssd/
```
dbkegley, line 41: We might be able to use this pattern to support `runAsCurrent` user with off-host execution. I don't think it would require any changes to Connect since we already assume uid/gid/passwd consistency between Connect and the content Pod.

colearendt: 😮 That's pretty cool!

dbkegley: @aronatkins pointed out that rsandbox/PAM password forwarding might pose a problem here, but it's worth a spike to see how far this gets us and if there are any other limitations.

dbkegley (approved):

PR #334, opened 2023-02-08
Replace connect team with individuals codeowners.

christierney:
These teammembers are also being added as admins to reduce the review/approve burden on Cole for simple changes (e.g. bumping release versions). Cole or Aaron should still be involved for more complicated changes.

@aronatkins @dbkegley @zackverham FYI: this came out of discussions on Trebuchet and work week. We want to be able to make small, well-understood updates, like bumping the Connect version, with less of a dependency outside the team. Please still be careful about approving other types of changes, like updating base images or other things that may break stuff.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,5 +5,5 @@
 /charts/rstudio-pm/* @rstudio/rspm-dev
 
 # connect resources
-/charts/rstudio-connect/* @rstudio/connect
-/examples/connect/* @rstudio/connect
+/charts/rstudio-connect/* @aronatkins @dbkegley @christierney @zackverham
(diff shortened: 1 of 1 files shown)


aronatkins (approved):
Send a PSA about this change to the Connect team once it lands.

colearendt (approved):
Also, just for completeness, it is important to keep in mind some of the important notions in this repo (and if they are not documented well, they should be... 🙈 WIP haha)

- if version number is not bumped on a chart change, it can cause issues with CI
- if the README is not updated properly, the README could point to the wrong chart version
- some issues in CI can cause the repo's index to get funky, which can force a complete rebuild of the repo's index
- CI from forks does not work fantastically 😞 

Good to be careful if CI is being weird before merging 😄

PR #335, opened 2023-02-08
`nodeSelector` consistency

pat-s:
#223 

We're running a mixed amd64/arm64 cluster lately and facing issues directing Posit pods to the amd64 instances due to the mixed `nodeSelector` implementation across the charts.

We can direct a bit via `affinity`, yet it does not always work (e.g. if the affinity nodes are at capacity and the arm64 node has space, the latter will be chosen). A consistent `nodeSelector` mapping would be more stable here.

Looking into the charts more in more detail, I think the issue for Connect specifically is the missing/wrong mapping of `nodeSelector` in `job.tpl`. Currently the values in `templateValues` are not passed along to the final configmap. And I couldn't infer where/how `placementConstraints` could be set.

(NB: I haven't looked into more detail for Workbench as currently our arm64 instances are too small anyway to be considered by the scheduler, but this will change at some point (and when we have finalized our arm64 image)).

I tried to align `nodeSelector` mappings across all charts, not sure though if all changes are enough/correct.

If there's a better way to approach this, just discard this effort :) Or make changes as needed.


colearendt:
Thanks for the PR! This is a great call! There will be a bit more that I will need to tie in here, I think, but this is definitely a great start! 

So the background here is that `placement-constraints` is a launcher feature that gets exposed as `nodeSelector` items. They are exposed directly by Workbench (so that users can select placement constraints interactively in the UI), but hidden / not implemented by Connect. Unfortunately it is also not possible to change the "default" placement constraints either. 

This seems like a nice workaround for situations such as yours when you can specify 

One quick question:
- let's say that you _did_ use placement constraints in Workbench (or they were implemented in Connect in the UI)
- would you expect that:
    - placement constraints from the application get _appended_ to a list of constraints provided / hard-coded by an administrator during the helm install
    - placement constraints from the application override the constraints provided by the helm install
    - the constraints provided by the helm install override the application selections?

I tend to lean towards the first option 😄

pat-s:
> So the background here is that placement-constraints is a launcher feature that gets exposed as nodeSelector items. They are exposed directly by Workbench (so that users can select placement constraints interactively in the UI),

Ah I see! Just checked it, did not have this on my radar so far. I wonder though: is this important for users? Aren't user only caring about getting a node that simply matches their resources?
Some examples to explain my thoughts better:

- If the user wants memory X, the user does not care whether the session is scheduled on a memory-optimized/GP/compute-optimized node. The admin cares as memory-optimized nodes would ofc save money if the memory request is high. 
- To make use of that feature, the user would need to understand what the different constraints imply. While this might be a nice feature for "expert users" I fear that exposing the option to users might cause more confusion on average than it helps (judging from the users of the environments I manage)

> let's say that you did use placement constraints in Workbench (or they were implemented in Connect in the UI)
would you expect that

Yes, option (1). The ones set on the admin side, i.e. in the helm chart, could be visible but locked for users and they can then append theirs (ahh, just trying it out in dev, I see this already applies! Maybe it could be differentiated visually a bit more.).
Yet I see also potential conflicts arising: what happens if the combination of admin and user constraint does not result in a valid configuration and the session won't be scheduled (e.g. let's say the admin specifies "amd64 only" and the user sets "compute-optimized" but the admin only defined a compute-optimized ASG for arm64) <- just a fictive example, but it should get the point.

One could argue that the admin needs to assert all possible combinations and that only valid ones are selectable by users. But this is probably not feasible in practice?

Another question that comes to my mind: how would one show users all available options? Maybe in a drop-down list? But then again, it's unclear which ones can be combined and which ones are mutually exclusive... 🤔

pat-s:
> but hidden / not implemented by Connect. 

Did I understand you right in that there is currently no other way than using `affinity` to get the node assignments for connect launcher pods right? I understood that for WB one can use `placementConstraints` for Connect does not have something similar? At least I couldn't find anything in the Admin guide.

> This seems like a nice workaround for situations such as yours when you can specify

This sentence reads a bit unfinished, maybe you wanted to add something more here?

pat-s:
@colearendt Cool to see this branch getting updated again.

I am using it for some time already internally and it solves our current nodeSelector issues.

Not sure if all proposed changes are 100% on point but feel free to make any changes! Happy to switch back to upstream once a solution got implemented (which I am seeing you're already working on judging from the last commits :) )

colearendt:
Sounds good!! Apologies if I end up breaking any of your changes. I think things should be backwards compatible with what you have done, but there may be some fiddling necessary 😄

charts/rstudio-connect/NEWS.md
```diff
@@ -1,3 +1,14 @@
+# 0.4.0
+
+- BREAKING: change `pod.nodeSelector` to `nodeSelector` for consistency with other charts
+  and the community. We error if this value is used to help highlight the change.
```
annikakrilov, line 4:
```suggestion
  and the community. We error if this value is used to help highlight the change.
```

This sentence could use some clarification, not sure what we're trying to say/convey here

PR #339, opened 2023-02-23
remove warnings related to support

colearendt:
Remove wording that is confusing to customers, and ensure folks understand how to make their production deployments reliable and reproducible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,10 @@ Supporting and miscellaneous charts with varying levels of maintenance, usefulne
 
 **IMPORTANT:**
 
-These charts are provided as a convenience to RStudio customers and are not formally supported by RStudio. If you
-have questions about these charts, you can ask them in the [issues](https://github.com/rstudio/helm/issues/new/choose)
-in the repository or to your support representative, who will route them appropriately.
+These charts are provided as a convenience to RStudio customers. If you have
+questions about these charts, you can ask them in the
+[issues](https://github.com/rstudio/helm/issues/new/choose) in the repository
+or to your support representative, who will route them appropriately.
 
 Bugs or feature requests should be opened in an [issue](https://github.com/rstudio/helm/issues/new/choose).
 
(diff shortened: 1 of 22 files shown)


andrie (commented):
LGTM

README.md
jthomasmock, line 1: Rather than Support did we want to say Customer Success?

jthomasmock (approved):
Left a comment re support vs CS.

README.md
andrie: I don't think so.  CS will only flag it to support, right?  Product installation issues seems like a support ticket domain to me.

jthomasmock: Makes sense, just wanted to make sure that was intentional. LGTM.

tylfin (approved):

charts/rstudio-library/README.md
```diff
@@ -1,18 +1,16 @@
...
+To ensure a stable production deployment, please:
 
-As a result, please:
 * Ensure you "pin" the version of the Helm chart that you are using. You can do
   this using the `helm dependency` command and the associated "Chart.lock" files
-  or the `--version` flag. IMPORTANT: This protects you from breaking changes
+  or the `--version` flag. **IMPORTANT: This protects you from breaking changes**
 * Before upgrading, to avoid breaking changes, use `helm diff upgrade` to check
   for breaking changes
 * Pay close attention to [`NEWS.md`](./NEWS.md) for updates on breaking
```
MariaSemple, line 16: Nit: Is it necessary to have this language repeated in so many places? Can it be added to a snippet that's pulled in or moved to a higher level to avoid having to update it in multiple places going forward?

MariaSemple (approved):
LGTM! Thanks Cole!

charts/rstudio-library/README.md
colearendt: Haha yes!! We actually do that 😄 If you notice the [charts/_templates.gotmpl](https://github.com/rstudio/helm/pull/339/files#diff-776f5ec4b6b88b8d9b2281357d65f5d8677508c334e70333182487cac0abe3ab) file, it is the "parent" of these children READMEs. `just docs` (after `just setup`) generates them, and we have CI that makes sure they are up-to-date a well. We still have to commit the files so that they can be viewed (so PRs still look gross) - it'd be nice if we could mark files as "generated" so they are lower priority in the PR review 😞

MariaSemple: Oh great, that's good to know!

PR #354, opened 2023-03-29
Update Workbench docs to clarify guidance on `config.profiles.profiles` vs. `config.server.profiles`

SamEdwardes:
## What

The docs currently advice users to configure `/etc/rstudio/profiles` using `config.profiles.profiles` in the values.yaml. From my understanding this is not correct:

- `/etc/rstudio/profiles` should be defined in `config.server.profiles`
- `/etc/rstudio/launcher.kubernetes.profiles.conf` should be defined in `config.profiles.launcher.kubernetes.profiles.conf`


colearendt (approved):
LGTM!

Issue #358, opened 2023-04-06
Service accounts in Connect hit a type error if the serviceaccountname is only numeric characters


zackverham:
To replicate in Connect: 

- Create a service account that only contains numeric characters (i.e. 1)
- Attach that service account to a piece of content
- Try to run the content

Connect UI shows a generic launcher failure, and the server logs show: 

```
time="2023-04-05T18:11:50.421Z" level=info msg="2023-04-05T18:11:50.421311Z [rstudio-kubernetes-launcher, log-source: Kubernetes] ERROR system error 71 (Protocol error) [description: Invalid status returned from create resource at /apis/batch/v1/namespaces/default/jobs: Bad Request - {\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"Job in version \\\"v1\\\" cannot be handled as a Job: json: cannot unmarshal number into Go struct field PodSpec.spec.template.spec.serviceAccountName of type string\",\"reason\":\"BadRequest\",\"code\":400}" stream=stderr subprocess=rstudio-launcher
time="2023-04-05T18:11:50.422Z" level=error msg="Unable to launch application: Launcher error: Post was not OK: Status code: 500" bundle_id=1 content_guid=d59d1599-f45a-48b5-a6a7-31269a180c88 content_id=1 mode=shiny worker=2964d24e
```

The issue is that the string here isn't quoted: 
https://github.com/rstudio/helm/blob/7ba0eb3840eb234ec0dfafadc8a8d75d55c411a8/examples/launcher-templates/default/2.3.0/job.tpl#L89

PR #359, opened 2023-04-06
 Fix quoting of service account in Connect's job template.

zackverham:
Bugfix for issue #358 

Enforces string quoting for service accounts in Connect's job template. This is needed for service accounts which only contain numeric characters. 

## QA Notes

Using the patched template - I'm able to now run a piece of content with a service account that uses a service account with only numeric characters: 

```
➜  kubectl get job run-shiny-application-rx29h -o yaml | grep "serviceAccountName"
      serviceAccountName: "1"
```

Double-checking for regressions - previously successful service accounts still work: 

```
➜  kubectl get job run-shiny-application-tz652 -o yaml | grep "serviceAccountName"
      serviceAccountName: another-sa
```


dbkegley:
We also need to update the helm templates. @colearendt since this change is backwards compatible, should we patch previous versions of the template, or should we create a new `2.3.0-v3` version and bump the rstudio-connect helm chart to the latest version?

https://github.com/rstudio/helm/tree/main/examples/launcher-templates/helm

atheriel (approved):

aronatkins (approved):

colearendt (approved):
LGTM! Let's merge this puppy!

PR #361, opened 2023-04-07
Add customImagesYaml docstring

dbkegley:


christierney (approved):

aronatkins (commented):
This is a doc-only change and doesn't need to adjust the version, yes? Merging should hold until after the Connect release.

dbkegley:
@aronatkins Unfortunately, we have to bump the chart version anytime we modify the values file

aronatkins:
I'm OK either way -- there's a slim chance that the release including these changes is not 2023.04.0, but we can reconcile if that ends up being the case. Your call.

zackverham (approved):

PR #363, opened 2023-04-13
Allow Launcher to set K8s API URL and Token from pod, Connect 2023.05, jammy

bschwedler:
This change is required to support the rotation of the service account token that is the default behavior in newer versions of Kubernetes.

The `prestart.bash` script for Connect was exporting
- `KUBERNETES_AUTH_TOKEN`
- `KUBERNETES_API_URL`

These were used by Connect to set entries in the configuration file.

When the token that was mounted in the container at startup would expire, all successive requests would fail with a `401` because it was using the auth token that was explicitly written to `launcher.kubernetes.conf`.

This PR will also serve as the RC for:
- Connect 2023.05.0 release
- `jammy` for the Connect image
- `jammy` for the content-init image
- `jammy` for the the content images in the default set of `Launcher.ClusterDefinitions`


colearendt (approved):
LGTM! Honestly, I think with a bit more investigation, we may be able to remove `prestart.bash` altogether 😮 This is great as a first pass though! Thanks!

colearendt (approved):
LGTM!

zackverham (approved):

nihara-thomas:
Merging these changes as part of the Connect 2023.05.0 release

PR #364, opened 2023-04-14
Bump workbench version and operating system

colearendt:
- change default operating system to jammy
- bump Workbench version to 2023.03
- add notes to the README to guide users


charts/rstudio-workbench/files/job.tpl
```diff
@@ -1,23 +1,34 @@
-# Version: 2
+# Version: 2.3.1
 # DO NOT MODIFY the "Version: " key
-# Helm Version: v3
+# Helm Version: v2
```
MariaSemple, line 3: Are we lowering the helm version?

PR #370, opened 2023-04-24
add a port to the service for metrics

colearendt:
This follows the pattern for other existing charts, and ensures that the ServiceMonitor we produce is actually functional.


aronatkins (approved):
CC @zackverham

zackverham (approved):

PR #371, opened 2023-04-27
updating default R version in PPM

shahmonanshi:


charts/rstudio-pm/README.md
```diff
@@ -116,7 +116,7 @@ The Helm `config` values are converted into the `rstudio-pm.gcfg` service config
 | awsAccessKeyId | bool | `false` | awsAccessKeyId is the access key id for s3 access, used also to gate file creation |
 | awsSecretAccessKey | string | `nil` | awsSecretAccessKey is the secret access key, needs to be filled if access_key_id is |
 | command | bool | `false` | command is the pod's run command. By default, it uses the container's default |
-| config | object | `{"HTTP":{"Listen":":4242"},"Metrics":{"Enabled":true},"Server":{"RVersion":"/opt/R/3.6.2/"}}` | config is a nested map of maps that generates the rstudio-pm.gcfg file |
+| config | object | `{"HTTP":{"Listen":":4242"},"Metrics":{"Enabled":true},"Server":{"RVersion":"/opt/R/4.2.0/bin/R"}}` | config is a nested map of maps that generates the rstudio-pm.gcfg file |
```
fh-mthomson, line 119: We also ran across this issue and set manually as done here; thanks for adding to the upstream chart!

Note: per the docs [here](https://docs.posit.co/rspm/admin/building-packages/?h=bin+r#r-configuration), the `bin/R` is not recommended
```suggestion
| config | object | `{"HTTP":{"Listen":":4242"},"Metrics":{"Enabled":true},"Server":{"RVersion":"/opt/R/4.2.0/"}}` | config is a nested map of maps that generates the rstudio-pm.gcfg file |
```

bschwedler: Use the default version (symlink) inside the container.
```suggestion
| config | object | `{"HTTP":{"Listen":":4242"},"Metrics":{"Enabled":true},"Server":{"RVersion":"/opt/R/default/"}}` | config is a nested map of maps that generates the rstudio-pm.gcfg file |
```

charts/rstudio-pm/values.yaml
```diff
@@ -228,7 +228,7 @@ serviceMonitor:
 # -- config is a nested map of maps that generates the rstudio-pm.gcfg file
 config:
   Server:
-    RVersion: "/opt/R/3.6.2/"
+    RVersion: "/opt/R/4.2.0/bin/R"
```
bschwedler, line 231: Use the default version (symlink) inside the container.
```suggestion
    RVersion: "/opt/R/default/"
```

charts/rstudio-pm/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: rstudio-pm
 description: Official Helm chart for RStudio Package Manager
-version: 0.5.7
+version: 0.5.8
```
bschwedler, line 3: Do you think we should update this to `0.6.0` since it could be a breaking change for some?

There would be two changes
1. R v3 is no longer installed by default
2. Rely on the default R version inside the container instead of pinning to a specific R version.

bschwedler:
Symlinks were added to the containers for the default versions of R & Python, which correspond to the image tags: https://github.com/rstudio/rstudio-docker-products/pull/601

bschwedler:
This has been superseded by #402

PR #397, opened 2023-07-21
Fix template env vars

colearendt:
Our template was not loading value-specified env vars properly


colearendt:
Going to go ahead and merge this since it is blocking some customer work. Please LMK if any issues - happy to follow up!

PR #402, opened 2023-07-31
Use R and Python image default paths

ianpittwood:


charts/rstudio-pm/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: rstudio-pm
 description: Official Helm chart for RStudio Package Manager
-version: 0.5.12
```
colearendt, line 3: Is this a breaking change? Otherwise bumping the minor version is probably not necessary?

examples/rbac/rstudio-launcher-rbac-0.2.17.yaml
colearendt, line 1: Weird that this file is showing up 😮  We must have messed up another PR somewhere

charts/rstudio-connect/ci/complex-values.yaml
colearendt, line 1: Wait, changing Connect here too?

colearendt (commented):
Generally LGTM - just a few questions about the scope of the change / effect on users.

charts/rstudio-pm/Chart.yaml
ianpittwood: I assumed it may be since we're pointing to a different R version path? But I can change it. I'll follow your guys' lead for what constitutes a breaking change.

examples/rbac/rstudio-launcher-rbac-0.2.17.yaml
ianpittwood: Yeah I didn't change that one :raised_eyebrow:

charts/rstudio-connect/ci/complex-values.yaml
ianpittwood: I changed the underlying base image to always include a /opt/<lang>/default symlink. Pretty much anywhere we're using a hardcoded Python/R path I want to change it to use that symlink. Since that file seemed to be just for CI stuff I thought I'd update it and that it wouldn't require a bump on the chart version. Let me know if that's incorrect though.

charts/rstudio-pm/Chart.yaml
bschwedler: It would be a breaking change if anybody is using the current image which does not have the symlink.

This was added in a PR that hasn't yet been merged into `main`
-  https://github.com/rstudio/rstudio-docker-products/pull/601

bschwedler (approved):
This would supersede #371 while also future-proofing against later version updates.

My only concern is making sure that this doesn't break with the existing `2023.04.0` images that are the default and _do not_ have the symlink.  We saw that pointing this to a non-existent R version results in a `CrashBackoffLoop`.

bschwedler (approved):
LGTM

SamEdwardes:
We can close this PR. A more flexible solution was merged via https://github.com/rstudio/helm/pull/474.

PR #404, opened 2023-08-07
add option to set subdir on the PVC volume mount

colearendt:
One thing that needs to be vetted: Does Launcher / the job launching process handle this subdir properly (and subdir below it?)... or will we need to pass this sub-directory along somehow?

Also need to:
- bump chart versions
- add NEWS entries


jforest (approved):
It looks good to me, and it would be great to get it out.

I'm happy to help test it in whatever way I can before I use it in my PR here: https://github.com/rstudio/aws-main-services/pull/625

colearendt:
Tested a bit and things look good!

PR #407, opened 2023-08-14
add pip.conf mount to /etc/pip.conf

### LINKED ISSUE #405: Workbench pip.conf configuration should automatically create a volume mount in the session pod
Currently, there isn't a canonical way for customers to define a pip.conf in our Workbench chart, such that it will be correctly picked up by pip in our session images. We should automate this process for customers, since many, if not the majority, will want to configure a configure pip.conf to define their internal pypi repos. This could be done under the `session.pip.conf` configuration and then we create a volumeMount to mount this file directly over /etc/pip.conf.

colearendt:
close #405 

Although we still need to think about Connect configuration for sessions too 🤷

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-workbench
 description: Official Helm chart for RStudio Workbench
-version: 0.6.5
+version: 0.6.6
 apiVersion: v2
 appVersion: 2023.06.0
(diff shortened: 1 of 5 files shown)


samcofer (approved):
Looks good to me!

samcofer:
Just did some additional testing here and I'm not seeing the volume mount in the pod for /etc/pip.conf. It's working in workbench, but not in the session pod.

samcofer:
Mistake on my part, missed enabling templating. Looks good and working as intended!

SamEdwardes:
Hey team - apologies I do not have time to review this before my time off!

PR #408, opened 2023-08-21
switch package manager default operating system to jammy / ubuntu2204

colearendt:


colearendt:
Worth noting that one big gotcha moving from bionic to jammy is that the default home directory permissions changed and no one outside of your user group can read anything in your home dir, even if you grant them read access to a subdirectory.

This should not be an issue here

PR #412, opened 2023-09-01
add terminationGracePeriodSeconds option

### LINKED ISSUE #411: no option for terminationGracePeriodSeconds
In a call w/ posit they told me that there is a issue where kubernetes might kill a pod before it can finish deactivating the license. To prevent that I should let the pod run for two minutes before killing it. I'm presuming to do that I would set `terminationGracePeriodSeconds`. However, there is no option to pass that into the helm values, and I don't use kustomize. Can that option be added in please?

colearendt:
close #411

### DIFF
--- a/charts/rstudio-connect/templates/deployment.yaml
+++ b/charts/rstudio-connect/templates/deployment.yaml
@@ -205,6 +205,9 @@ spec:
       {{- if .Values.pod.sidecar }}
 {{ toYaml .Values.pod.sidecar | indent 6 }}
       {{- end }}
+      {{- with .Values.pod.terminationGracePeriodSeconds }}
+      terminationGracePeriodSeconds: {{ . }}
+      {{- end }}
       volumes:
       - name: rstudio-connect-config
         configMap:
--- a/charts/rstudio-connect/Chart.yaml
+++ b/charts/rstudio-connect/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-connect
 description: Official Helm chart for RStudio Connect
-version: 0.5.5
+version: 0.5.6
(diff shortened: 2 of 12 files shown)


aronatkins (approved):

charts/rstudio-connect/templates/deployment.yaml
```diff
@@ -204,6 +204,9 @@ spec:
       {{- end }}
       {{- if .Values.pod.sidecar }}
 {{ toYaml .Values.pod.sidecar | indent 6 }}
+      {{- end }}
+      {{- with .Values.pod.terminationGracePeriodSeconds }}
+      terminationGracePeriodSeconds: {{ . }}
       {{- end }}
```
AlexMapley, line 210:
```suggestion
      terminationGracePeriodSeconds: {{  .Values.pod.terminationGracePeriodSeconds }}
```
What do you think about just setting this value directly? We set a default integer value of `120` for `terminationGracePeriodSeconds` in all of our charts, but here we're handling it here as if it might be empty or `None`.

AlexMapley (commented):
Sorry for the driveby review, I wasn't request on this PR. But I noticed this branch and realized I should be adding the same option to the chronicle chart - https://github.com/rstudio/helm/pull/443

charts/rstudio-connect/templates/deployment.yaml
AlexMapley: This is how this looks in the chronicle chart: https://github.com/rstudio/helm/pull/443/files#diff-95db5aca0ede6323eb40498fe2af5fc708cac66d5603bc813e4f979139d7a88cR66

PR #417, opened 2023-09-20
Remove CRAN override from Connect chart value

fh-mthomson:
Motivation: If a user specifies a non-PPM version of `CRAN` (e.g., `https://cran.r-project.org/`)  when deploying to Connect, the current Helm chart overrides it with the PPM version of CRAN, which may not yet have the latest version of a package.

This change (1) removes this constraint (while still having `RSPM` pointing to PPM as a fall back) and (2) mirrors the `workbench` Helm chart (changed [here](https://github.com/rstudio/helm/commit/19b2d2f86c3b836d767114e633895697b36e5b2f#diff-923d1824addd75f58b6022881e5a99553726a4e1c18c97e03bdf025b5f6ab9a5L391))

### DIFF
--- a/charts/rstudio-connect/Chart.yaml
+++ b/charts/rstudio-connect/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-connect
 description: Official Helm chart for RStudio Connect
-version: 0.5.6
+version: 0.5.7
 apiVersion: v2
 appVersion: 2023.09.0
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
--- a/charts/rstudio-connect/NEWS.md
+++ b/charts/rstudio-connect/NEWS.md
@@ -1,3 +1,7 @@
+# 0.5.7
+
+- Remove `CRAN` as a default repository entry in `session.repos.conf`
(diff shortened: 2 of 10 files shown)


dbkegley:
[Connect `v2023.09`](https://docs.posit.co/connect/news/index.html#new) has some modifications to the default PackageURL behavior. 

> Connect has new strategies for resolving the set of R package repositories used to install packages in content environments, and controlling how configured RPackageRepository entries and content repositories are used. These are controlled with the setting R.PackageRepositoryResolution. The default behavior has changed. The new default, `lax`, uses a combined list of configured repositories and repositories from content. To use only configured repositories and discard content repositories, set to `strict`. To use the old overriding behavior, set to `legacy`. See the [Admin Guide](https://docs.posit.co/connect/admin/r/package-management/#configuring-on-posit-connect) for more details and examples. (#23853)

Does the new default behavior make this change unnecessary?

cc @toph-allen

fh-mthomson:
Brilliant! `lax` is exactly what I think would solve this. Much appreciated!

For more context, see https://support.posit.co/hc/en-us/requests/91166 where we'd previously discussed.

PR #421, opened 2023-09-28
Bump workbench version to 2023.09.0

melissa-barca:
- bump workbench version
- Reconfigure load-balancing so balancer is not explicitly set to `sessions`, as this is now the default behavior, and to set `delete-node-exit` so that Workbench manages the nodes in the database.


atheriel (approved):
LGTM

melissa-barca:
@atheriel is it expected for the PR status checks to take this long?

colearendt:
Oh sorry for delay - we can merge this. 

GItHub Actions prevents checks from running on GitHub Actions commits (it updated the docs). This is to prevent a runaway loop of commits + actions.

PR #441, opened 2023-11-15
Add the Chronicle Server chart

AlexMapley:
## Background
Part of https://github.com/rstudio/chronicle/issues/683.

We are migrating this from a private posit repo, where we've been hosting the chart previously.

## Summary
- Migrates the chronicle chart to this repo
- Updates with rstudio helm best practices, standard features
- Update the global file `charts/_templates.gotmpl`
- Adds ci components and documentation

## Notes
This PR is still an early draft, though I'd like to get some early feedback if possible and open the PR now for discussion

I've added a number of new features to the chronicle chart like pod tolerations, affinities, nodeSelectors, etc. - though i didn't add every feature under the sun. I'm on the lookout for more basic features we think the chronicle chart should include


charts/_templates.gotmpl
```diff
@@ -8,6 +8,9 @@ R and Python packages that Data Scientists need to create and share data product
   {{- else if eq .Name "rstudio-workbench" -}}
 Data Scientists use [RStudio Workbench](https://www.rstudio.com/products/workbench/) to analyze data and create data
 products using R and Python.
+  {{- else if eq .Name "posit-chronicle" -}}
+IT Administrators and Business Users use [Posit Chronicle](https://www.rstudio.com/products/chronicle/) to aggregate and monitor
```
AlexMapley, line 12: This url doesn't exist yet and https://www.rstudio.com/products/chronicle will 404, this is a stub for now until we release a suitable doc link

bdeitte: I just happened to check this PR out as I saw the message, neat to see!  A nitpick here, but I don't think this will be at "products".  Perhaps have it mimic pro-drivers as the closest example?  https://docs.posit.co/chronicle

Also we should switch these all to posit.co!

charts/posit-chronicle/LICENSE
```diff
@@ -0,0 +1,21 @@
+# MIT License
+
+Copyright (c) 2023 Posit PBC
```
bdeitte, line 3: Oh I check one PR out and I'm just full of nitpicks.  It's Posit Software PBC

charts/_templates.gotmpl
AlexMapley: yeah i was thinking about this, I don't think this really belongs as a product either so maybe I can leave this template alone. https://docs.posit.co/chronicle seems like a good future link too!

> Also we should switch these all to posit.co!
Yeah I definitely agree. I'm hoping at least we can set the right standards with chronicle to help guide the name migration of our other charts - but then chronicle will kindof stick out in the interim.

charts/posit-chronicle/LICENSE
AlexMapley: Nice catch! We've had this incorrectly copyrighted to `Posit PBC` in the private chart repo too 😐

charts/_templates.gotmpl
```diff
@@ -8,6 +8,9 @@ R and Python packages that Data Scientists need to create and share data product
   {{- else if eq .Name "rstudio-workbench" -}}
 Data Scientists use [RStudio Workbench](https://www.rstudio.com/products/workbench/) to analyze data and create data
 products using R and Python.
+  {{- else if eq .Name "posit-chronicle" -}}
+IT Administrators and Business Users use Posit Chronicle](https://www.rstudio.com/chronicle/) to aggregate and monitor
```
t-margheim, line 12:
```suggestion
IT Administrators and Business Users use [Posit Chronicle](https://www.rstudio.com/chronicle/) to aggregate and monitor
```

charts/posit-chronicle/README.md
```diff
@@ -0,0 +1,204 @@
+# Posit Chronicle
+
+![Version: 0.1.0](https://img.shields.io/badge/Version-0.1.0-informational?style=flat-square) ![AppVersion: 2023.11.0](https://img.shields.io/badge/AppVersion-2023.11.0-informational?style=flat-square)
+
+#### _Helm chart for the Chronicle Server_
+
+IT Administrators and Business Users use Posit Chronicle](https://www.rstudio.com/chronicle/) to aggregate and monitor
```
t-margheim, line 7:
```suggestion
IT Administrators and Business Users use [Posit Chronicle](https://www.rstudio.com/chronicle/) to aggregate and monitor
```

charts/posit-chronicle/README.md
```diff
@@ -0,0 +1,204 @@
...
+helm search repo rstudio/posit-chronicle -l
+```
+
+## Usage
+
+The Chronicle chart is meant to be used in tandem with other workbench and connect instances.
+To enable the chronicle agent, additional values will have to be passed to your
+workbench and connect values.
+
+Here's some example agent helm values to run the agent sidecar in workbench:
```
t-margheim, line 42:
```suggestion
The Chronicle chart is meant to be used in tandem with other Workbench and Connect instances.
To enable the Chronicle agent, additional values will have to be passed to your
Workbench and Connect values.

Here's some example agent helm values to run the agent sidecar in Workbench:
```

charts/posit-chronicle/README.md
```diff
@@ -0,0 +1,204 @@
...
+      - name: logs
+        mountPath: "/var/lib/rstudio-server/audit"
+      env:
+      - name: CHRONICLE_SERVER_ADDRESS
+        value: "http://chronicle-server.default.svc.cluster.local"
+...
+```
+
+And here's some example agent helm values for connect, where we utilize an api key stored as a k8s secret
+to scrape the connect rest server:
```
t-margheim, line 62:
```suggestion
And here's some example agent helm values for Connect, where we utilize an API key stored as a k8s secret
to scrape the Connect metrics REST service:
```

charts/posit-chronicle/README.md
```diff
@@ -0,0 +1,204 @@
...
+      - name: CHRONICLE_SERVER_ADDRESS
+        value: "http://chronicle-server.default.svc.cluster.local"
+      - name: CONNECT_API_KEY
+        valueFrom:
+          secretKeyRef:
+            name: connect
+            key: apikey
+```
+
+Note that it will be up to the user to provision that connect api key Kubernetes secret.
```
t-margheim, line 82:
```suggestion
Note that it will be up to the user to provision that Connect API key Kubernetes secret.
```

charts/posit-chronicle/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+``` yaml
+config:
+  localStorage:
+    enabled: true
+    location: "./chronicle-data"
+    retentionPeriod: "30d"
+```
+
+RetentionPeriod accepts a duration string input. 0 implies infinite retention, disabling file expiration.
```
t-margheim, line 100:
```suggestion
`RetentionPeriod` accepts a duration string input. `0` implies infinite retention, disabling file expiration.
```

charts/posit-chronicle/README.md
```diff
@@ -0,0 +1,204 @@
...
+For example:
+1s for 1 second, 5m for 5 minutes, 12h for 12 hours, 7d for one week, 365d for one year, 0 for unbound retention.
```
t-margheim, line 102:
```suggestion
`1s` for 1 second, `5m` for 5 minutes, `12h` for 12 hours, `7d` for one week, `365d` for one year, `0` for unbound retention.
```

charts/posit-chronicle/README.md
```diff
@@ -0,0 +1,204 @@
...
+  annotations: {
+    eks.amazonaws.com/role-arn:  arn:aws:iam::123456789000:role/iam-role-name-here
+  }
+  # -- Additional labels to add to the chronicle-server serviceaccount
+  labels: {}
+```
+
+If you are unable to use IAM Roles for Service Accounts, there are any number of
+alternatives for injecting AWS credentials into a container. As a fallback, the
+s3 storage config allows specifying a profile:
```
t-margheim, line 140:
```suggestion
S3 storage config allows specifying a profile:
```

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+
+{{ template "rstudio.install" . }}
+
+## Usage
+
+The Chronicle chart is meant to be used in tandem with other workbench and connect instances.
+To enable the chronicle agent, additional values will have to be passed to your
+workbench and connect values.
+
+Here's some example agent helm values to run the agent sidecar in workbench:
```
t-margheim, line 17:
```suggestion
The Chronicle chart is meant to be used in tandem with other Workbench and Connect instances.
To enable the Chronicle agent, additional values will have to be passed to your
Workbench and Connect values.

Here's some example agent helm values to run the agent sidecar in Workbench:
```

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+      - name: logs
+        mountPath: "/var/lib/rstudio-server/audit"
+      env:
+      - name: CHRONICLE_SERVER_ADDRESS
+        value: "http://chronicle-server.default.svc.cluster.local"
+...
+```
+
+And here's some example agent helm values for connect, where we utilize an api key stored as a k8s secret
+to scrape the connect rest server:
```
t-margheim, line 37:
```suggestion
And here's some example agent helm values for Connect, where we utilize an API key stored as a k8s secret
to scrape the Connect metrics REST service:
```

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+  localStorage:
+    enabled: true
+    location: "./chronicle-data"
+    retentionPeriod: "30d"
+```
+
+RetentionPeriod accepts a duration string input. 0 implies infinite retention, disabling file expiration.
+For example:
+1s for 1 second, 5m for 5 minutes, 12h for 12 hours, 7d for one week, 365d for one year, 0 for unbound retention.
+Units shorter than seconds or longer than days, such as milliseconds and weeks, are not supported.
```
t-margheim, line 78:
```suggestion
`RetentionPeriod` accepts a duration string input. `0` implies infinite retention, disabling file expiration.
For example:
`1s` for 1 second, `5m` for 5 minutes, `12h` for 12 hours, `7d` for one week, `365d` for one year, `0` for unbound retention.
Units shorter than seconds or longer than days, such as milliseconds and weeks, are not supported.
```

charts/posit-chronicle/README.md
AlexMapley: wait i think the helm linter actually just reverted this update with a followup commit: https://github.com/rstudio/helm/pull/441/commits/70826ea346ad8a46e7eed13df8614999a5e6c658... hm..

AlexMapley: i think the linter was actually confused on some newline spacing, https://github.com/rstudio/helm/pull/441/commits/a8b77ea0f8291af66bb871442bab562f31ece826 should fix

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,85 @@
+image:
+  repository: "935931255537.dkr.ecr.us-east-2.amazonaws.com/chronicle"
```
atheriel, line 2: We should be using the public image now:

```suggestion
  repository: "ghcr.io/rstudio/chronicle"
```

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+  tag: "latest"
+  imagePullPolicy: "Always"
+
+serviceaccount:
+  enabled: false
```
atheriel, line 7: Nit: should we use `create` here instead to be more consistent with the other charts?

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+  localStorage:
+    enabled: true
+    location: "./chronicle-data"
+    retentionPeriod: "30d"
+  s3Storage:
+    enabled: false
+    bucket: "posit-chronicle"
+    prefix: ""
+    profile: ""
+    region: "us-east-2"
```
atheriel, line 72: I don't think it's very intuitive to set a default region here, we may want to leave this blank instead.

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+  https:
+    # If https.enabled=true, we will ignore any http
+    # values and enable https in our config instead
+    enabled: false
+    listen: ":443"
+    key: ""
+    certificate: ""
+  localStorage:
+    enabled: true
+    location: "./chronicle-data"
```
atheriel, line 65: I don't see a reason to allow admins to set the mount path inside the container, so I think we should remove this option and have the generated config/StatefulSet default to something like `/data`.

```suggestion
```

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+service:
+  rest:
+    # -- The port to use for the REST service
+    port: 80
+    # -- The port to forward REST requests to on the pod. Also see pod.port
+    targetPort: 5252
+    # -- Additional annotations to add to the chronicle-server service
+    annotations: {}
+    # -- Additional labels to add to the chronicle-server service
+    labels: {}
```
atheriel, line 25: There's no reason users should be able to configure the internal port mapping, we should just hardcode it to `80:5252` like we do for the Package Manager chart and remove a lot of potential footguns:

```suggestion
service:
  # -- Additional annotations to add to the chronicle-server service
  annotations: {}
  # -- Additional labels to add to the chronicle-server service
  labels: {}
```

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+# -- The number of replica pods to maintain for this service
+replicas: 1
+
+pod:
+  command: "/chronicle"
+  args: ["start", "-c", "/opt/chronicle/config.gcfg"]
+  rest:
+    port: 5252
```
atheriel, line 34: As mentioned above: I'd suggest removing these.

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+  persistentVolumeSize: 1Gi
+
+config:
+  http:
+    listen: ":5252"
+  https:
+    # If https.enabled=true, we will ignore any http
+    # values and enable https in our config instead
+    enabled: false
+    listen: ":443"
```
atheriel, line 60: As mentioned elsewhere, I'd suggest we remove any configuration of the exact ports themselves and focus on whether this is HTTP/HTTPS only:

```suggestion
  https:
    # -- When true, listen for HTTPS connections rather than HTTP.
    enabled: false
```

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,85 @@
...
+    region: "us-east-2"
+    compactionEnabled: false
+  metrics:
+    enabled: true
+  profiling:
+    enabled: false
+    listen: :3030
+  tracing:
+    enabled: false
+    address: ""
```
atheriel, line 81: Since there's no facility for getting this data out right now and these are hidden in the default config, I'd suggest removing these entirely so they don't get picked up by customers copy & pasting this `values.yaml` file (which happens regularly).

```suggestion
```

charts/posit-chronicle/README.md
AlexMapley: ah, i was editing this file directly not the template, go figure 😅.
Got it fixed now!

charts/posit-chronicle/Chart.yaml
```diff
@@ -0,0 +1,23 @@
...
+annotations:
+  artifacthub.io/license: MIT
+  artifacthub.io/links: |
+    - name: Posit Community
+      url: https://community.rstudio.com/c/r-admin/5
+    - name: About Posit Team
+      url: https://posit.co/products/enterprise/team/
+keywords:
+  - "rstudio"
+  - "chronicle"
```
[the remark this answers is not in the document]
AlexMapley: I like this too, though i wasn't sure on my own if we wanted to adhere to these `rstudio` keywords for other reasons. If you're onboard though let's go for it

charts/posit-chronicle/LICENSE
```diff
@@ -0,0 +1,21 @@
+# MIT License
+
+Copyright (c) 2023 Posit Software PBC
```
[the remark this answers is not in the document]
AlexMapley: ah yes you are right, that lines up with `rstudio/rstudio` too at https://github.com/rstudio/rstudio/blob/6d2346880ec61544c71b238057d1de9cad8f99a7/src/cpp/server/ServerMetrics.cpp#L4.

charts/posit-chronicle/values.yaml
AlexMapley: Nice! yes let's try this out, i'd forgotten this was public now.

charts/posit-chronicle/values.yaml
AlexMapley: I'm definitely for to removing the tracing portion, the port definitions i'm not sold on yet but will consolidate to another thread

charts/posit-chronicle/values.yaml
AlexMapley: sure we can do that, consistency should be the main goal and on reflection i do think `serviceaccount.create` is more clear than `serviceaccount.enable`.

`serviceaccount.createAndUse` might be most accurate, but that's a little much and not in line with our other charts.

charts/posit-chronicle/values.yaml
AlexMapley: I do like removing the `rest` section indentation, i'm going to address that in a followup commit now.
I also agree with you that in practice, a user rarely needs to alter the default ports. We might be able to get away with this without overly constraining our users, especially if we aren't using `hostPorts` which might conflict with the underlying k8s node.

Generally though if we can configure these values in our chronicle config, it makes sense that we allow configuring them in our helm config. As we are now a user can only configure a chronicle server in helm with a subset of configurations they could do locally, which doesn't feel too great.

I might even say, instead of removing these port definitions lets actually make this values file _more verbose_ so that we capture 100% of possible values a user can set in their chronicle server config locally. I know that sounds dogmatic, but on the other hand why would we have certain config settings that are immutable for Kubernetes deployments?

charts/posit-chronicle/values.yaml
AlexMapley: I'm removing all the port definitions as we discussed in meeting 👍

i think we should keep the profiling bit though, i've actually made use of pprof a few times in we'll want to continue using it for our longer term memory utilization QA

charts/posit-chronicle/values.yaml
AlexMapley: I do kindof like the extra documentation of having this at the top level of our helm values, volume mappings can be a little tricky to grok sometimes just from the code.
It's not needed though I agree

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,78 @@
...
+    enabled: false
+    bucket: "posit-chronicle"
+    prefix: ""
+    profile: ""
+    region: ""
+    compactionEnabled: false
+  metrics:
+    enabled: true
+  profiling:
+    enabled: false
```
AlexMapley, line 74: @atheriel I realized that i've been using pprof in our k8s environment, and plan on continuing to do so for some future QA. I would like to kep this!

AlexMapley: here's an example of me doing this:
```shell
 % kubectl port-forward chronicle-server-0 3030:3030
Forwarding from 127.0.0.1:3030 -> 3030
Forwarding from [::1]:3030 -> 3030
Handling connection for 3030
```

Screenshot 2023-11-21 at 12 30 32 PM

charts/posit-chronicle/values.yaml
AlexMapley: Also we'll reuse this value in a couple places, it's also nice to just be able to reference `config.localStorage.location` when we do. That argument could have been valid for our port definitions we've since hardcoded too, though i'm fine with setting those directly

t-margheim (approved):

charts/posit-chronicle/values.yaml
atheriel: Great! I didn't realize it was useful in its current form. Happy to keep it in then.

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+
+``` yaml
+serviceaccount:
+  enabled: false
+  # -- Additional annotations to add to the chronicle-server serviceaccount
+  annotations: {
+    eks.amazonaws.com/role-arn:  arn:aws:iam::123456789000:role/iam-role-name-here
+  }
+  # -- Additional labels to add to the chronicle-server serviceaccount
+  labels: {}
```
atheriel, line 111: I'd suggest we use a more concise example. Also drop the `{}` and use more Kubernetes-like YAML formatting:

```suggestion
  enabled: true
  annotations:
    eks.amazonaws.com/role-arn: "arn:aws:iam::123456789000:role/iam-role-name-here"
```

AlexMapley:
@colearendt @shahmonanshi what do you 2 think of this chart in its current form? 
Are there others you might think we'd benefit from tagging as reviewers? I wasn't sure who else to add.

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+
+{{ template "rstudio.install" . }}
+
+## Usage
+
+The Chronicle chart is meant to be used in tandem with other Workbench and Connect instances.
+To enable the Chronicle agent, additional values will have to be passed to your
+Workbench and Connect values.
+
+Here's some example agent helm values to run the agent sidecar in Workbench:
```
atheriel, line 17:
```suggestion
Here are some example Helm values to run the agent sidecar in Workbench:
```

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+      - name: logs
+        mountPath: "/var/lib/rstudio-server/audit"
+      env:
+      - name: CHRONICLE_SERVER_ADDRESS
+        value: "http://chronicle-server.default.svc.cluster.local"
+...
+```
+
+And here's some example agent helm values for Connect, where we utilize an API key stored as a k8s secret
+to scrape the Connect metrics REST service:
```
atheriel, line 37:
```suggestion
And here are some example Helm values for Connect, where we also pass a Connect
API key from a Kubernetes Secret to unlock more detailed metrics:
```

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+            name: connect
+            key: apikey
+```
+
+Note that it will be up to the user to provision that Connect API key Kubernetes secret.
+
+## Storage Configuration
+
+Chronicle can be configured to store its data in a local kubernetes volime, in S3,
+or in both.
```
atheriel, line 62:
```suggestion
Chronicle can be configured to persist data to a local Kubernetes volume, AWS
S3, or both.
```

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+
+The default configuration uses a local volume, which is suitable if you'd like to 
+access and analyze the data within your cluster:
+
+``` yaml
+config:
+  localStorage:
+    enabled: true
+    location: "./chronicle-data"
```
atheriel, line 71: Nit: I'd prefer an absolute path for containers.
```suggestion
    location: "/chronicle-data"
```

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+`RetentionPeriod` accepts a duration string input. `0` implies infinite retention, disabling file expiration.
+For example:
+`1s` for 1 second, `5m` for 5 minutes, `12h` for 12 hours, `7d` for one week, `365d` for one year, `0` for unbound retention.
+Units shorter than seconds or longer than days, such as milliseconds and weeks, are not supported.
+
+
+You can disable local storage by setting `localStorage.enabled` to `false`, and you can enable S3
+storage by setting `S3Storage.enabled` to `true`. Enabling both is also acceptable,
+and the server will store to both places. When using S3, you must also set the `S3Storage.Bucket`
+parameter, like so:
```
atheriel, line 84:
```suggestion
`retentionPeriod` controls how long usage data are kept. For example, `"120m"`
for 120 minutes, `"36h"` for 36 hours, `14d` for two weeks, or `"0"` for
unbounded retention. (Units smaller than seconds or larger than days are not
supported.)

You can also persist data to AWS S3 instead of (or in addition to) local
storage:
```

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+
+``` yaml
+config:
+  s3Storage:
+    enabled: true
+    bucket: "posit-chronicle"
+    prefix: ""
+    profile: ""
```
atheriel, line 92: Nit: don't bother showing empty fields.
```suggestion
```

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+If you are unable to use IAM Roles for Service Accounts, there are any number of
+alternatives for injecting AWS credentials into a container. As a fallback,
+the S3 storage config allows specifying a profile:
+
+``` yaml
+config:
+  s3Storage:
+    enabled: true
+    bucket: "posit-chronicle"
+    prefix: ""
```
atheriel, line 123:
```suggestion
```

charts/posit-chronicle/templates/stateful-set.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - name: chronicle-server
+        image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+        imagePullPolicy: {{ .Values.image.imagePullPolicy }}
+        command:
+        - {{ .Values.pod.command }}
+        {{- if .Values.pod.args }}
+        args:
+        {{- toYaml .Values.pod.args | nindent 8 }}
+        {{- end }}
+        ports:
```
atheriel, line 46: Are we missing the profiling port?

charts/posit-chronicle/ci/simple-values.yaml
```diff
@@ -0,0 +1,25 @@
+image:
+  tag: "latest"
+
+serviceaccount:
+  enabled: false
```
atheriel, line 5: Right?

```suggestion
  create: true
```

charts/posit-chronicle/ci/no-local-storage-values.yaml
```diff
@@ -0,0 +1,21 @@
+image:
+  tag: "latest"
+
+serviceaccount:
+  enabled: false
```
atheriel, line 5: Right?

```suggestion
  create: true
```

charts/posit-chronicle/templates/stateful-set.yaml
AlexMapley: i did a double take on this too, technically we don't need to expose it when we'd be accessing pprof using `kubectl port-forward`

charts/posit-chronicle/ci/no-local-storage-values.yaml
AlexMapley: nice catch, looks our ci runs aren't exactly all encompassing. i'll double check this locally as well,

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,131 @@
...
+        value: "http://chronicle-server.default.svc.cluster.local"
+...
+```
+
+And here's some example agent helm values for Connect, where we utilize an API key stored as a k8s secret
+to scrape the Connect metrics REST service:
+
+```yaml
+pod:
+  ...
```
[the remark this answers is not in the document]
AlexMapley: I'll remove this now, you're right we technically only need to present the a subset of values here to be correct.

charts/posit-chronicle/README.md.gotmpl
AlexMapley: whoops forgot to commit this suggestion, I'm doin so now.

atheriel (approved):

colearendt (approved):
Just a few nits. Totally fine to update at a later date 😄 I try to avoid Statefulsets if possible, but if you have a good reason to use them (slower to roll over, etc. etc.), then carry on!

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,80 @@
+image:
+  repository: "ghcr.io/rstudio/chronicle"
+  tag: "latest"
```
tnederlof, line 3: Similar comment about using the latest tag. Our other charts use the appVersion, could that make sense here too?

charts/posit-chronicle/values.yaml
AlexMapley: Oh sorry I forgot to follow up on this, I tested and found without this value the server would crash - so it actually is essential.
Might something we might want to improve in the server configuration layer.

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,120 @@
...
+Workbench or Connect server pods by setting `pod.sidecar` in their respective
+`values.yaml` files.
+
+Here are some example Helm values to run the agent sidecar in Workbench:
+
+```yaml
+pod:
+  sidecar:
+    - name: chronicle-agent
+      image: posit-chronicle:latest
```
tnederlof, line 25: It feels like putting the latest tag in the example might not be a great idea, could we template it to use the `appVersion` instead like our other helm charts?

charts/posit-chronicle/README.md.gotmpl
```diff
@@ -0,0 +1,120 @@
...
+```
+
+And here are some example Helm values for Connect, where we also pass a Connect
+API key from a Kubernetes Secret to unlock more detailed metrics:
+
+```yaml
+pod:
+  sidecar:
+    - name: chronicle-agent
+      image: posit-chronicle:latest
```
tnederlof, line 43: It feels like putting the latest tag in the example might not be a great idea, could we template it to use the `appVersion` instead like our other helm charts?

tnederlof (approved):
Looks good, just some comments about using the latest tag.

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,80 @@
...
+  # -- Additional selector labels to add to the chronicle-server pods
+  selectorLabels: {}
+  # -- A map used verbatim as the pod's "affinity" definition
+  affinity: {}
+  # -- A map used verbatim as the pod's "nodeSelector" definition
+  nodeSelector: {}
+  # -- An array used verbatim as the pod's "tolerations" definition
+  tolerations: []
+
+# We do want at least 1Gi of persistent volume space
```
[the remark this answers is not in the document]
AlexMapley: Can do! These are the same comments from our old chart, where the audience has been internal

charts/posit-chronicle/values.yaml
AlexMapley: Can do! We can use [v2023.11.2](https://github.com/rstudio/chronicle/pkgs/container/chronicle) which we're publishing now.
That's a good callout, we were just talking on our team about which image to use

charts/posit-chronicle/README.md.gotmpl
AlexMapley: We can totally pin this to a non-lates image 👍 . Weversion the agent and server on a separate cadence, but probably can enforce this 1:1 versioning as a policy.
We've been thinking of migrating them both into 1 codebase, that would probably help too. This is one more reason to do that

charts/posit-chronicle/Chart.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - https://github.com/rstudio/helm
+maintainers:
+  - name: sol-eng
+    email: docker@rstudio.com
+    url: https://github.com/rstudio/helm
+annotations:
+  artifacthub.io/license: MIT
+  artifacthub.io/links: |
+    - name: About Posit Team
+      url: https://posit.co/products/enterprise/team/
```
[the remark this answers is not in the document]
AlexMapley: This was a discussion on our team, we weren't sure about linking posit community if there wasn't any guarantee there'd be chronicle related content there yet. I think it was @atheriel and I discussing this

But I think it could be nice adding them, maybe we can get some chronicle community engagement going in `r-admin`. 

The Posit Teams adds is also great, chronicle probably will run best it's current form as a 1:1 per posit team setup.

charts/posit-chronicle/ci/complex-values.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+config:
+  https:
+    # If https.enabled=true, we will ignore any http
+    # values and enable https in our config instead
+    enabled: true
+    key: /path/to/key
+    certificate: /path/to/cert
+  localStorage:
+    enabled: true
```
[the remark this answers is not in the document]
AlexMapley: In the chart values we enable local storage and not s3 storage by default, this is us testing out both in one of our values files we run `helm template` against in our ci

charts/_templates.gotmpl
```diff
@@ -25,6 +29,8 @@ products using R and Python.
 # RStudio Package Manager
   {{- else if eq .Name "rstudio-workbench" -}}
 # RStudio Workbench
+  {{- else if eq .Name "posit-chronicle" -}}
+# Posit Chronicle
```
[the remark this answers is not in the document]
SamEdwardes: Yeah, it is a tough call. Either way, it is a bit awkward. I think I would vote for calling it `posit-chronicle` as you have since that it the actual product name. Hopefully, in the future, we can update the other charts in a non-breaking way.

charts/posit-chronicle/ci/complex-values.yaml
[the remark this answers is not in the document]
SamEdwardes: Ughh sorry - I thought I was reviewing the default values.yaml file when I was looking at this. My apologies.

charts/posit-chronicle/ci/complex-values.yaml
```diff
@@ -0,0 +1,65 @@
...
+    test-label: test-service-label
+    test-label2: test-service-label2
+  selectorLabels:
+    test-selector-label: test-service-selectorLabel
+    test-selector-label2: test-service-selectorLabel2
+
+storage:
+  persistentVolumeSize: 10Gi
+
+config:
```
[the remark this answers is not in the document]
AlexMapley: nice callout. I noticed we don't do this in the [workbench-config section](https://github.com/rstudio/helm/blob/main/charts/rstudio-workbench/values.yaml#L371), but if ppm and connect do we might as well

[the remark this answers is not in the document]
AlexMapley: Definitely a worthwhile callout. We've done a lot of work already to pare down this config file, and went even further within this PR migrating from https://github.com/rstudio/platform-helm/blob/main/charts/chronicle/values.yaml, so this will likely continue as we get more comfortable with our default settings

charts/posit-chronicle/templates/stateful-set.yaml
```diff
@@ -0,0 +1,80 @@
+---
+apiVersion: apps/v1
+kind: StatefulSet
```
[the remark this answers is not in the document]
AlexMapley: It's mostly as a convenience for the pvc provisioning, if you're running with local storage enabled Persistent storage a pvc is a must-have for most use cases. It's nice to not have to worry about duplicate replica sets being created on helm updates as well, the pod replacement policy of stateful-sets is generally simple and non-problematic for chronicle.

Theoretically we could run chronicle on a deployment/replica-set model, there's not too many technical blockers to think of. The stateful-set model has worked well so far though, and our local storage situation will be a bit trickier without it.

charts/posit-chronicle/values.yaml
```diff
@@ -0,0 +1,80 @@
...
+  # -- Additional selector labels to add to the chronicle-server service
+  selectorLabels: {}
+
+# -- The number of replica pods to maintain for this service
+replicas: 1
+
+pod:
+  # -- The command and args to run in the chronicle-server container
+  command: "/chronicle"
+  args: ["start", "-c", "/opt/chronicle/config.gcfg"]
```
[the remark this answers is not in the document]
AlexMapley: That's a great idea too, let's add this in a followup commit next

charts/posit-chronicle/ci/complex-values.yaml
```diff
@@ -0,0 +1,65 @@
...
+    key: /path/to/key
+    certificate: /path/to/cert
+  localStorage:
+    enabled: true
+    location: "./chronicle-data"
+    retentionPeriod: "5m"
+  s3Storage:
+    enabled: true
+    bucket: "posit-chronicle"
+    region: "us-east-2"
```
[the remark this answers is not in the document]
AlexMapley: we have to or the chronicle server will barf, also this is for our ci testing too

charts/posit-chronicle/values.yaml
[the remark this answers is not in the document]
SamEdwardes: Thanks Alex! For context, Connect and Package Manager only have 1 config file (although with many more options than Chronicle). They include the product name in the filename and in the directory name. This is repetitive, but I just want to call out wherever we can be consistent across different products!

Workbench has many configuration files.

charts/posit-chronicle/ci/complex-values.yaml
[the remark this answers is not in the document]
SamEdwardes: Workbench is a little bit different, but I would use Connect and Package Manager as my reference since they are newer products, and they are pretty consistent between each other in terms of naming conventions.

charts/posit-chronicle/templates/stateful-set.yaml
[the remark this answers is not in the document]
colearendt: That makes sense.

It's worth noting that statefulsets do have some downsides: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#limitations

Usually services that uses statefulsets have a particular reason to do so. Otherwise replicasets/deployments are the more standard choice.

If you use a PVC, does each chronicle node store its own data? Or do they share data storage?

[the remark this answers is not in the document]
AlexMapley: That's tough, we do have a choice to make for sure. As a developer I feel like it's nice to have those pvcs deleted by default:

> Deleting and/or scaling a StatefulSet down will not delete the volumes associated with the StatefulSet. This is done to ensure data safety, which is generally more valuable than an automatic purge of all related StatefulSet resources.

Though on the other hand, we are trying to build a data archive too. If we scaled a chronicle server from 1->5 replicas for a day, and stored a bunch of data, it would be nice if 2 weeks later when we tried scaling 1->5 again we found our old data in those local storage systems that we could add to. 

> If you use a PVC, does each chronicle node store its own data? Or do they share data storage?

Yeah each chronicle server does need a dedicated pvc, that's a known technical limitation right now in the alpha though. We need to rewrite some in-memory locks into file-locks for our compactor engine, and then we can support nfs or shared storage. We'll definitely need to make some helm changes around that too.

charts/posit-chronicle/values.yaml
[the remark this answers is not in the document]
AlexMapley: You know what you're right, i'm renaming this `/etc/posit-chronicle/posit-chronicle.gcfg` as you suggest. We should be consistent now and think about refactoring later.

AlexMapley:
Thanks everyone for all for your feedback! I don't think I disagree with anything, I tried to make 80-85% of the adjustments suggested and made a note of the rest with some github issue tracking.

- Most notably from @SamEdwardes, we're now using pascal case for our config values and following other shared config conventions. 
- @tnederlof your comments pushed us to make a new release, and we now tag against the `2023.11.2` docker image
- We have a [new epic ](https://github.com/rstudio/chronicle/issues/715) around your nfs comments @colearendt, we've definitely been thinking about supporting nfs too post this launch

PR #444, opened 2023-12-05
Update licenses, add chart keyword and update descriptions

AlexMapley:
## Background
RStudio has rebranded to Posit, and over time we're migrating our code to reflect that. 
This PR is a small step towards that migration for our helm charts, and won't require any change on the users behalf.

## Summary
- Update helm chart licenses
- Update chart descriptions to reference Posit products
- Adds the `posit` keyword to all helm charts for better search-ability


shepherdjerred (approved):

aronatkins (approved):

kfeinauer (approved):

AlexMapley:
This PR is still completely valid and i'd like to merge. I don't think we should bump up the chart version for these changes, but we're going to continue to fail our `Lint and Test Charts` step until we do. I'd like to bypass branch protections for this merge instead, and leave our chart versions unchanged.

I run all of these charts in our platform EKS environment, when I see a new version comes out I'll typically upgrade to new chart versions just to see what new things I get or issues are solved. I would be disappointed as a user if I upgraded to a new chart version, just to find the chart metadata changed and nothing else.

PR #453, opened 2024-01-09
Workbench: Make `prestart-launcher.bash` distribution-agnostic

pat-s:
(Assuming no other distros are being used yet as a base for WB images.)

Right now, the script is hard-coded for debian-like distros only.
The following patch adds a dynamic solution to also include rhel-like distros.

Should be pretty fail-safe, but feel free to modify to your liking, of course.


colearendt:
Love this! Thanks for the submission!

PR #474, opened 2024-03-01
PPM - set default for `config.Server.RVersion` and `config.Server.PythonVersion` to `""`

### LINKED ISSUE #473: Update the default Package Manager config to take advantage of R and Python autodetection
The current Package Manager docker image and helm chart have hard-coded R and Python configurations.

## Related

- https://github.com/rstudio/rstudio-docker-products/issues/523

## The issue

**Docker**

https://github.com/rstudio/rstudio-docker-products/blob/e0fa3ab39e64e6e9b03febcbcb97e71a96210596/package-manager/rstudio-pm.gcfg#L12-L14

```docker
; Git sources require a configured R installation. R is often installed at `/usr/lib/R`
; or `/usr/lib64/R`.
RVersion = /opt/R/default/
```

**Helm**

```yaml
config:
  Server:
    RVersion: "/opt/R/3.6.2/"
  HTTP:
    Listen: :4242
  Metrics:
    Enabled: true
```

Hard coding these values can lead to errors when starting Package Manager. In particular, this is an issue with hard coding the value into Helm because there is no guarantee that `/opt/R/3.6.2/` is installed.

## Proposed fix

- Update the default Helm configuration to use R and Python auto-detection.
- There is also an opportunity to simplify the Package Manager docker image, but that is out of the scope of this repo.

SamEdwardes:
## Description

Update the default values `config.Server.RVersion` and `config.Server.PythonVersion to `""`. This will ensure that Package Manager uses R and Python autodetection unless the user explicitly sets the desired R or Python version in the Helm chart.

## Related

- Closes #473 
- I think we could also close this PR: https://github.com/rstudio/helm/pull/402

## Testing

I did a test deployment of the changes. I confirmed that autodetection kicked in as expected:

```logs
2024/03/01 22:34:21 Checking for newly supported operating systems and R versions
2024/03/01 22:34:21 Updated supported operating systems and R versions to latest
2024/03/01 22:34:21 Info: User namespace sandbox test successful.
2024/03/01 22:34:21 `Server.RVersion` was not set. Attempting to auto-detect.
2024/03/01 22:34:22 R 4.2.3 auto-detected on the system successfully at /opt/R/4.2.3/bin/R. To set the version manually configure the `Server.RVersion` option.
2024/03/01 22:34:22 `Server.PythonVersion` was not set. Attempting to auto-detect.
2024/03/01 22:34:22 Python 3.10.12 auto-detected on the system successfully at /usr/bin/python3. To set the version manually configure the `Server.PythonVersion` option.
2024/03/01 22:34:22 Warning: Unable to build Git packages for Python on this system: required Python modules `build`, `virtualenv` are not installed
2024/03/01 22:34:22 Posit Package Manager service is ready
```


charts/rstudio-workbench/test-values.yaml
```diff
@@ -0,0 +1,17 @@
+# Configures user home directory shared storage
```
tnederlof, line 1: @SamEdwardes, it looks like you committed your test file.

charts/rstudio-pm/values.yaml
```diff
@@ -232,7 +232,10 @@ serviceMonitor:
 # -- config is a nested map of maps that generates the rstudio-pm.gcfg file
 config:
   Server:
-    RVersion: "/opt/R/3.6.2/"
```
tnederlof, line 235: Since these are the product defaults I feel like removing them entirely would be appropriate. I understand having them explicit for the transition away from RVersion though so perhaps we leave them explicitly set to blank for a while then remove entirely to keep the values clean?

tnederlof (changes requested):

tylfin:
FYI we have a fix for this in the upcoming release:

> Warning: Unable to build Git packages for Python on this system: required Python modules `build`, `virtualenv` are not installed

https://github.com/rstudio/package-manager/pull/13097

charts/rstudio-workbench/test-values.yaml
SamEdwardes: Ooos sorry about that! I will fix.

charts/rstudio-pm/values.yaml
SamEdwardes: Good point on removing them entirely. 

I think we should just remove it entirely in this PR. The change is captured in the changelog, so we don't need to also document the change in the values.yaml.

I will make the update.

SamEdwardes:
@tnederlof thank you for the review. I made the suggested changes. Regarding what config options to document I removed R and Python from the config since we can rely on the product's default behaviour.

tnederlof (approved):

shepherdjerred (approved):

PR #478, opened 2024-03-06
Change Connect default service.type and other service generation items

tnederlof:
This PR contains the following changes, following the changes made to RSPM by @atheriel in 2021 (PR here: https://github.com/rstudio/helm/pull/122) This change to RSPM was made over 2 years ago and went pretty smoothly, now bringing it to Connect.

- Use a `ClusterIP` service by default instead of a `NodePort`.
- Add support for controlling [`loadBalancerIP`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) and [`clusterIP`](https://kubernetes.io/docs/concepts/services-networking/service/#choosing-your-own-ip-address), which are commonly-used service features.
- Ignore nodePort settings when the service type is not `NodePort`. Otherwise, this will generate server-side validation errors.
- Simplify handling of Service annotations by removing the existing YAML helper is not actually necessary and has a confusing name.
- Tweak documentation for `service.nodePort` and `service.annotations`.

Since the first of these is a breaking change, I bumped the minor version for the chart.

## Moving to ClusterIP

`ClusterIP` is the Kubernetes default, exposing the service in-cluster only. This is a secure, well-understood default. For external users, there are a few options:

- Use an `Ingress`. We support this already and it works with `ClusterIP`.
- Use a `LoadBalancer`. This can be expensive so it makes a poor default.
- Use a `NodePort` and some kind of external load balancing solution.

It seems like the last route (which is the current default) is much more uncommon than the first two, so I'd advocate for this change.

In addition:

- `NodePorts` are a very limited cluster resource, so we should avoid consuming them if we can avoid it.
- Defaulting to `NodePort` is a poor security choice, especially if we have users that don't understand the consequences.

### DIFF
--- a/charts/rstudio-connect/Chart.yaml
+++ b/charts/rstudio-connect/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-connect
 description: Official Helm chart for RStudio Connect
-version: 0.5.14
+version: 0.6.0
 apiVersion: v2
 appVersion: 2024.02.0
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
--- a/charts/rstudio-connect/NEWS.md
+++ b/charts/rstudio-connect/NEWS.md
@@ -1,6 +1,6 @@
 # 0.6.0
 
-- BREAKING: The generated service will now have type `ClusterIP` by default.
+- BREAKING: The generated service will now have type `ClusterIP` set by default.
 - Add support for setting the `loadBalancerIP` or `clusterIP`.
 - Ignore `nodePort` settings when the service is not a `NodePort`.
 - Improve the documentation for some service-related settings.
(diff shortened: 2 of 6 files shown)


dbkegley (approved):
Long overdue change, thanks @tnederlof !

dbkegley:
I installed the helm chart from this branch using our local development setup for Connect and everything worked as expected 👍 

```
david@Davids-MacBook-Pro:connect [main] » k get svc
NAME                      TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)           AGE
connect-rstudio-connect   ClusterIP   10.109.240.7   <none>        80/TCP,9108/TCP   6s
```

tnederlof:
@dbkegley thanks, works on my side and I verified annotations still make it through on all types. I am unable to merge bc I think CI doesn't like the latest commit being an auto one. If you are able to when the appropriate people have reviewed please do!

atheriel (approved):

PR #485, opened 2024-03-29
Create documentation site

edavidaja:
This PR creates a Quarto documentation site with fuller examples, currently deployed in preview at:
 
https://connect.posit.it/drafts/docs-posit-co/helm/

Some further changes to the templating and CI workflows will be required.


dbkegley:
This is awesome, thanks for getting this started. I'm really happy to see us moving to quarto for these docs.

A big chunk of these helm docs is the templated table that enumerates all the possible values and defaults. This table is really tough to read with the current styling. It would be nice if we could make it a little easier to read. I know this is probably part of what you're referring to by "some further changes to the templating...". Let me know if you're looking for ideas on how to make the README.md and README.md.tpl work together. 

This is getting even more complicated as we now have markdown and quarto and go-templates in the mix.

Screenshot 2024-04-01 at 12 24 30 PM

Our Connect docs have a similar problem for one of our tables. We've added some styling to make the table scrollable and the sidebar gets automatically collapsed so it doesn't get displayed over top:

```
:::{.column-screen-inset-right style="overflow-x: auto;"}
```

Screenshot 2024-04-01 at 12 24 16 PM

edavidaja:
> I know this is probably part of what you're referring to by "some further changes to the templating...". Let me know if you're looking for ideas on how to make the README.md and README.md.tpl work together.

It is, and I am. I think it may involve some combination of:

- [render the readmes from the templates in a pre-render script](https://quarto.org/docs/projects/scripts.html#pre-and-post-render)
- [render some content conditionally based on whether it's targeted to `gfm` or `html`](https://quarto.org/docs/authoring/conditional.html)
- [getting different table output out of helm-docs](https://github.com/norwoodj/helm-docs?tab=readme-ov-file#advanced-table-rendering)

colearendt (approved):
LGTM!

Issue #463, opened 2024-02-15
Add quarto to default Connect config


tnederlof:
Quarto is present in the image but it is not enabled in the config

PR #486, opened 2024-04-26
connect: update content images and enable quarto by default

tnederlof:
- Adds in the new content-base and content-pro images in the `default-runtime.yaml` and `default-runtime-pro.yaml` files created via this PR: https://github.com/rstudio/rstudio-docker-products/pull/737
- Enables `Quarto` and `Python` in the default `values.yaml` `config` section so Python and Quarto content will work out of the box for local execution or off-host execution (closes https://github.com/rstudio/helm/issues/463). Previously Python was automatically enabled only when launcher was enabled. This conditional is removed now that its in the values.

## Testing

Deployed Connect into EKS using base and pro, they show up as expected and content can be deployed from R and Python (tested on the content-pro:r4.2.3-py3.11.9-ubuntu2204 image).

Base env list:
env list base

Pro env list:
env list pro

### DIFF
--- a/charts/rstudio-connect/values.yaml
+++ b/charts/rstudio-connect/values.yaml
@@ -335,6 +335,8 @@ config:
     DataDir: /var/lib/rstudio-connect
   Scheduler:
     InitTimeout: 5m
+  Quarto:
+    Enabled: true
   Logging:
     ServiceLog: STDOUT
     ServiceLogLevel: INFO    # INFO, WARNING or ERROR
--- a/charts/rstudio-connect/values.yaml
+++ b/charts/rstudio-connect/values.yaml
@@ -335,8 +335,6 @@ config:
     DataDir: /var/lib/rstudio-connect
   Scheduler:
     InitTimeout: 5m
-  Quarto:
-    Enabled: true
   Logging:
     ServiceLog: STDOUT
     ServiceLogLevel: INFO    # INFO, WARNING or ERROR
@@ -349,3 +347,5 @@ config:
     GraphiteHost: 127.0.0.1
     GraphitePort: 9109
     GraphiteClientId: rsconnect
+  # Note: Quarto and Python are both enabled automatically when launcher.enabled: true
+  # https://github.com/rstudio/helm/blob/main/charts/rstudio-connect/templates/_helpers.tpl
--- a/charts/rstudio-connect/values.yaml
+++ b/charts/rstudio-connect/values.yaml
@@ -333,6 +333,12 @@ config:
   Server:
     Address: http://localhost:3939
     DataDir: /var/lib/rstudio-connect
+  Python:
+    Enabled: true
+    Executable: "/opt/python/3.9.17/bin/python3"
+  Quarto:
+    Enabled: true
+    Executable: "/opt/quarto/1.3.340/bin/quarto"
   Scheduler:
     InitTimeout: 5m
   Logging:
--- a/charts/rstudio-connect/values.yaml
+++ b/charts/rstudio-connect/values.yaml
@@ -335,7 +335,9 @@ config:
     DataDir: /var/lib/rstudio-connect
   Python:
     Enabled: true
-    Executable: "/opt/python/3.9.17/bin/python3"
+    Executable:
+      - /opt/python/3.9.17/bin/python
+      - /opt/python/3.8.17/bin/python
   Quarto:
     Enabled: true
     Executable: "/opt/quarto/1.3.340/bin/quarto"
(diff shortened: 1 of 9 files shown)


tnederlof:
@aronatkins, I think enabling Quarto by default in the config is not breaking anything, but I wanted to check with you. Without it the environments all show Quarto as None by default.

charts/rstudio-connect/values.yaml
```diff
@@ -335,6 +335,8 @@ config:
     DataDir: /var/lib/rstudio-connect
   Scheduler:
     InitTimeout: 5m
+  Quarto:
```
aronatkins, line 338: We should configure a Quarto interpreter location for when this chart is used with local execution. The current images should have `/opt/quarto/1.3.340/bin/quarto`, but that will shift to 1.4.X after the Connect 2024.04.0 release.

We should also consider `Python.Enable=true`. Unfortunately, if this chart is used with local execution, that means that we need to configure at least one Python interpreter. If Python is enabled and there are no interpreters, Connect will fail to start.

CC @dbkegley - do you have a different opinion? I'm surprised that the chart doesn't enable Python already!

aronatkins: NEWS should call out that Quarto is enabled.

aronatkins (approved):

aronatkins:
Quarto is currently different from R/Python in that it can be enabled without a configured executable. At some point, we probably want all three R/Python/Quarto to behave more similarly (warn if enabled without executable).

charts/rstudio-connect/values.yaml
tnederlof: Yes, I did add the news. I know in PPM, putting in a direct path caused issues, as if it isn't updated, it will break. Can you add to the sequence for when Connect version is bumped to make sure the quarto path is checked?

dbkegley: Yeah, I think python not being enabled for local execution by default is just an artifact of this chart being heavily influenced by off-host. Since python is already installed in the default Connect image I don't see any problem enabling it for local execution by default, that seems like the right call.

tnederlof: The latest commit leaves the default values alone now and only enabled Quarto like Python when launcher is enabled. Maybe worth a separate PR if we want to consider turning those on for the none OHE case.

tnederlof:
@dbkegley I am nervous to put in hardcoded paths for Python and Quarto for non OHE since they need to be updated. When it was in PPM like that and they didn't get updated when the images changed, the charts would break. If the Connect team has a process to update when the images changed then it would make sense but want to make sure that is in place first.

dbkegley:
> If the Connect team has a process to update when the images changed then it would make sense but want to make sure that is in place first.

Appreciate the caution. I'll create a followup for the Connect team to find a way to flag this. I think having an GHA in this repo would be a nice addition to our pre-release testing

tnederlof:
@dbkegley Okay I made the changes, now setting Python and Quarto directly in the default config section of the values.yaml and removed the conditional launcher only logic. Tested both ways on my side and things work as expected. Mind giving it a look on your side and when good merge?

charts/rstudio-connect/values.yaml
```diff
@@ -333,6 +333,12 @@ config:
   Server:
     Address: http://localhost:3939
     DataDir: /var/lib/rstudio-connect
+  Python:
+    Enabled: true
+    Executable: "/opt/python/3.9.17/bin/python3"
```
dbkegley, line 338: There are 2 python installations available on the default image, should we include both?

```suggestion
    Executable:
      - /opt/python/3.9.17/bin/python
      - /opt/python/3.8.17/bin/python
```

dbkegley:
> One thing I wanted to check on is whether the Executables defined for Quarto and Python are ignored in OHE (is there any harm in having them in there in the values?). Or do you think its cleaner to still have conditional logic where when in OHE we omit those Executable lines?

Good point. If OHE is enabled then any executables defined in the default `values.yaml` will be ignored so everything should work as expected, but it might confuse folks. Maybe we should add a comment in the values noting that these are the defaults for local execution only and the defaults for off-host are defined in the default runtimes.yaml

tnederlof:
@dbkegley I changed the comments in values.yaml to make it clear that the executables do not apply for OHE and linked to the docs. I retested and things look good to me. I think this is ready when you want to merge (not sure who else should take a look).

dbkegley (approved):
LGTM, I think this is good to merge anytime.

I'll be updating the quarto version during the next Connect release which is happening.. soon :tm:  so we'll get some mileage out of the new GHA right away. The new checks should fail when we update the image during the release.

PR #487, opened 2024-04-30
Connect: Enable Quarto and Python by default for local execution

dbkegley:
~~This PR enables Quarto and Python by default for local execution. They were already enabled by default for off-host execution.~~ This change will happen in the [original PR ](https://github.com/rstudio/helm/pull/486) instead. This PR only add the new GHA

Adds a new GHA test that runs on every PR which validates `that` the local defaults defined in the values are actually installed on the default image:
- Enumerates the interpreter versions that are defined in the helm chart's default values for each R, Python, and Quarto.
- Then attempts to run `command -v` against each executable to ensure that it exists in the default chart image.

Related to https://github.com/rstudio/helm/pull/486#issuecomment-2085390506

### DIFF
--- a/.github/workflows/chart-test.yaml
+++ b/.github/workflows/chart-test.yaml
@@ -84,3 +84,22 @@ jobs:
         if: ${{ github.ref == 'refs/heads/main' }}
         run: ct install --target-branch main --all --chart-dirs charts --chart-dirs other-charts
         continue-on-error: true
+
+  check-versions-connect:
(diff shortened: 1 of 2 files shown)


aronatkins (approved):
#️⃣ 🔢 

Something that I found confusing: There is also `charts/rstudio-connect/Makefile`. Some product-specific targets are in the Justfile and others are within the various Makefile.

PR #491, opened 2024-05-02
Use the full PWB version docker tag

cm421:
The tag for PWB image `rstudio/rstudio-workbench:ubuntu2204-2024.04.0` doesn't exist on docker hub for some reason, which causes problems because that's the tag used in PWB's helm chart.

While platform team investigates why that tag wasn't created, we can use `rstudio/rstudio-workbench:ubuntu2204-024.04.0-735.pro3` instead

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-workbench
 description: Official Helm chart for RStudio Workbench
-version: 0.7.3
+version: 0.7.4
 apiVersion: v2
 appVersion: 2024.04.0
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -18,9 +18,9 @@ dependencies:
 annotations:
   artifacthub.io/images: |
     - name: rstudio-workbench
-      image: rstudio/rstudio-workbench:ubuntu2204-2024.04.0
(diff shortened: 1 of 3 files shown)


kfeinauer (approved):

kfeinauer:
@cm421 Need to bump the chart version before merge.

tnederlof:
I think those annotations are only for artifacthub.io, the values under `image` and `session.image` are what drive the images pulled by default. The chart is set up to take the app version + tagPrefix and pull automatically, to override we would need to set the following two tags with `ubuntu2204-2024.04.0-735.pro3`:

`image.tag` https://github.com/rstudio/helm/blob/90bc753151c4bdbb96d16cdaf505ad914f7b7dd4/charts/rstudio-workbench/values.yaml#L141
`session.image.tag` https://github.com/rstudio/helm/blob/90bc753151c4bdbb96d16cdaf505ad914f7b7dd4/charts/rstudio-workbench/values.yaml#L29

cm421:
So changing `appVersion: 2024.04.0` to `appVersion: 2024.04.0+735.pro3` in charts/rstudio-workbench/Chart.yaml should set the default `tag` as well?

Anyone with more helm experience know if changing that will affect other areas as well?

tnederlof:
> So changing `appVersion: 2024.04.0` to `appVersion: 2024.04.0+735.pro3` in charts/rstudio-workbench/Chart.yaml should set the default `tag` as well?
> 
> Anyone with more helm experience know if changing that will affect other areas as well?

It would apply to the image, but it looks like the variable is used in a couple of spots. Since all the helm charts are consistent across products in this area, it would be nice to leave it as is if the image can be corrected on Monday.

bdeitte:
Given the many branch conflicts and how old this is now, closing this out.

Issue #422, opened 2023-10-04
Pass database passwords as env variables from secrets


tnederlof:
As discussed in https://github.com/rstudio/helm/issues/420 it would be ideal to manage database passwords as Kubernetes secrets and have the helm chart inject them as env variables into the correct container.

To accomplish this, Workbench will need to allow passwords to be set from env variables: https://github.com/rstudio/rstudio-pro/issues/5024

Connect and Package Manager already have the functionality.
https://docs.posit.co/connect/admin/appendix/configuration/#environment-variables
https://docs.posit.co/rspm/admin/appendix/configuration/#environment-variables

edavidaja:
This is supported and documented in each product:

Workbench:
- https://docs.posit.co/ide/server-pro/admin/database/configuration.html#environment-variables
- https://docs.posit.co/helm/charts/rstudio-workbench/README.html#database-password

Connect:
- https://docs.posit.co/connect/admin/appendix/configuration/#environment-variables
- https://docs.posit.co/helm/charts/rstudio-connect/README.html#database-password

Package Manager:
- https://docs.posit.co/rspm/admin/appendix/configuration.html#environment-variables
- https://docs.posit.co/helm/charts/rstudio-pm/README.html#database

PR #493, opened 2024-05-06
Provide helper to set DB password as env variables across products

tnederlof:
This PR closes: https://github.com/rstudio/helm/issues/422 by providing a consistent way to set PostgreSQL database passwords as env variables from K8s secrets in Workbench, Connect and Package Manager. This is now possible since Workbench supports setting the [database password as an env variable](https://docs.posit.co/ide/server-pro/database/configuration.html#environment-variables) in the 2024.04.0 release (Connect and Package Manager already supported this). Currently, admins need to either set the password in the values or imperatively during `helm install`. Both are not very k8s native ways of doing things.

`rstudio-library` contains a `_database-env.tpl` helper to create an env variable from a secret. This is then used by each of the product charts.

For each product chart, the following values can now be specified:
```
database:
  password:
    secret: posit-database-secret
    secretKey: password
```

Package Manager has an additional set of values for the usage database since it is separate:
```
database:
  usageDataPassword:
    secret: posit-database-secret
    secretKey: password
```

This PR also includes a common README template. After this is live, we can change the admin guides to use these options, streamlining the install process.


SamEdwardes:
This is a great idea. I have not done a code review, but I agree with the change in spirit. A few initial comments:

- Connect has an optional metrics database config: https://docs.posit.co/connect/admin/database/postgres/#operational-metrics. Did we cover this?
- Would it be simpler and easier for us to maintain if we just documented an example of how to set the DB password as an environment variable from a secret?

tnederlof:
> This is a great idea. I have not done a code review, but I agree with the change in spirit. A few initial comments:
> 
> * Connect has an optional metrics database config: https://docs.posit.co/connect/admin/database/postgres/#operational-metrics. Did we cover this?
> * Would it be simpler and easier for us to maintain if we just documented an example of how to set the DB password as an environment variable from a secret?

Thanks for taking a look Sam. Good catch on the Connect side, will add those settings (optional if someone wants to use a separate db/schema).

It's a good question about whether should we reduce complexity and instead of using rstudio-library helper (this PR) just document in each chart setting the following which will accomplish the same thing.
```
pod:
  env:
    - name: WORKBENCH_POSTGRES_PASSWORD
      valueFrom:
        secretKeyRef:
          name: secret-name
          key: secret-key
```

I would love thoughts from the different product teams. If that approach seems more straightforward can refactor this to a docs things only.

tnederlof:
I am going to close this one and go with a similar approach proposed by @SamEdwardes. After talking with a few folks it seems a more straightforward path that accomplishes the same thing, with the added benefit of showing folks how to set env variables in our helm charts.

PR #496, opened 2024-05-06
Fix: Unset the default [Server].Address option for PPM

tylfin:


jmwoliver (approved):

charts/rstudio-pm/NEWS.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.5.24
+
+- Unset the [Server].Address default value as this can cause issues when serving from a domain name.
```
glin, line 5:
```suggestion
- Unset the `[Server].Address` default value as this can cause issues when serving from a domain name.
```
Or `config.Server.Address` which was the format used in previous NEWS items

glin (approved):

calebAtIspot:
>  as this can cause issues when serving from a domain name

What issues does this cause specifically?

PR #498, opened 2024-05-06
Bump PPM version to 2024.04.0-20

nodivbyzero:


glin (approved):

glin:
Looking at the linter failure, it looks like it fails because we're changing the NEWS for an existing chart version. Anyone know if this is okay to merge? Version `0.5.23` is really on build `-20` now, so the NEWS isn't accurate anymore. Or we could just leave the NEWS as-is, since there's not much difference between `-18` and `-20` besides a PPM release notes bug fix.

PR #503, opened 2024-05-10
run helm-docs as quarto pre-render

edavidaja:
- **updated posit-docs extension**
- **run helm-docs as quarto pre-render**


edavidaja:
superseded by #504

PR #504, opened 2024-05-14
Editing pass - Phase 1

AshleyHenry15:


edavidaja:
I think if the READMEs for the products are generated during a `quarto render` from dev, I think we can eliminate the `chart-doc` action on PRs entirely.

colearendt (approved):
LGTM!

Issue #424, opened 2023-10-16
Inconsistent behavior across charts


Almenon:
The posit connect chart specifies the following:

```
    {{- if eq .Values.strategy.type "RollingUpdate" }}
    rollingUpdate:
      maxUnavailable: {{ .Values.strategy.rollingUpdate.maxUnavailable }}
      maxSurge: {{ .Values.strategy.rollingUpdate.maxSurge }}
    {{- end }}
```

The PM chart just has:
```
  {{- with .Values.strategy }}
  strategy:
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

It would be nice if the templating was consistent, as this tripped me up when working with the charts. I thought just specifying strategy.type: RollingUpdate was enough because it worked for the connect chart, but in the PM chart you also need to specify `rollingUpdate: null` under `strategy`.

colearendt:
Thanks for reporting @Almenon ! To be clear: do you think that the Connect chart's approach is preferable?

Almenon:
I don't have much of a preference, but if you want me to pick one as the 'best' I'll pick the PM chart as it is more flexible. But up to you :)

PR #508, opened 2024-05-19
[PM] Fix deployment strategy

calebAtIspot:
fixes https://github.com/rstudio/helm/issues/424

It also fixes a bug where ArgoCD is unable to update the deployment when the strategy type is recreate. 

```
error when patching "/dev/shm/2215103101": Deployment.apps "rstudio-packagemanager-rstudio-pm" is invalid: 

spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'
```


calebAtIspot:
Also if you're wondering why I don't just pass in `null` for the `rollingUpdate` value when I call the chart, it's because of https://github.com/helm/helm/issues/12637

tnederlof:
Thank you @calebAtIspot I went through the changes and think they make sense as this would bring PPM inline with Workbench and Connect's charts and the default values are for `RollingUpdate` so the desired default should still work. @atheriel

atheriel (approved):
LGTM. Thank you for the contribution!

PR #510, opened 2024-05-20
update actions

edavidaja:
- silence some node version warnings
- update some unsupported versions of python

FYI: chart-releaser no longer supports the `charts-repo` parameter.
It wasn't clear to me how to update that in a manner consistent with
releasing charts to helm.rstudio.com so for now I've left that in place.


atheriel (approved):
:sunglasses:

PR #514, opened 2024-05-22
workbench: Add database config and pass as env variable to README

tnederlof:
This PR relates to https://github.com/rstudio/helm/issues/422 by providing customers guidance on how to configure and set database passwords consistently and in a Kubernetes-friendly way across products.

The docs added show how to configure the database and set up PostgreSQL database passwords as env variables from K8s secrets in Workbench. This is now possible since Workbench supports setting the [database password as an env variable](https://docs.posit.co/ide/server-pro/database/configuration.html#environment-variables) in the 2024.04.0 release. Similar changes are being submitted for each of the products.

Currently, admins need to either set the password in the values or imperatively during helm install. Both are not very k8s native ways of doing things.

This PR is already inflight for Admin install doc changes: https://github.com/rstudio/rstudio-pro/pull/6316

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-workbench
 description: Official Helm chart for Posit Workbench
-version: 0.7.4
+version: 0.7.5
 apiVersion: v2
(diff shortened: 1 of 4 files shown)


atheriel (approved):
LGTM

PR #515, opened 2024-05-22
connect: Add database config and password as env to README

tnederlof:
This PR relates to https://github.com/rstudio/helm/issues/422 by providing customers guidance on how to configure and set database passwords consistently and in a Kubernetes-friendly way across products.

The docs added show how to configure the database and set up PostgreSQL database passwords as env variables from K8s secrets in Connect. Similar changes are being submitted for each of the products.

Currently, admins need to either set the password in the values or imperatively during helm install. Both are not very k8s native ways of doing things.

This PR is already in flight for Admin install doc changes: https://github.com/rstudio/connect/pull/26928


tnederlof:
@nihara-thomas fyi this is the companion to this change in the admin guide: https://github.com/rstudio/connect/pull/26928

charts/rstudio-connect/README.md.gotmpl
```diff
@@ -24,6 +24,46 @@ To function, this chart requires the following:
 
 {{ template "rstudio.licensing" . }}
 
+## Database
+
+Connect requires a PostgreSQL database when running in Kubernetes. You must configure a [valid connection URI and a password](https://docs.posit.co/connect/admin/database/postgres/) for the product to function correctly. Both the connection URI and password may be specified in the `config` section of `values.yaml`. However, we recommend only adding the connection URI and putting the database password in a Kubernetes `Secret`, which can be [automatically set as an environment variable](#database-password).
```
dbkegley, line 29: nit: I don't think Postgres is strictly required, but it is definitely preferred. Feel free to ignore this suggestion if we want to push users toward postgres as the default.

```suggestion
If you are using a PostgreSQL database for Connect when running in Kubernetes, then you must configure a [valid connection URI and a password](https://docs.posit.co/connect/admin/database/postgres/). Both the connection URI and password may be specified in the `config` section of `values.yaml`. However, we recommend only adding the connection URI and putting the database password in a Kubernetes `Secret`, which can be [automatically set as an environment variable](#database-password).
```

dbkegley (approved):

charts/rstudio-connect/README.md.gotmpl
tnederlof: I think at some point I heard we wanted to just require Postgres since customers often miss it and in almost all cases they need it (unless they are running Connect on a single node without NFS storage is the only case I can think of). It is in the requirements currently: https://docs.posit.co/connect/admin/getting-started/off-host-install/#feature-requirements

That being said its not my call so what ever the Connect team decides I am cool with.

dbkegley: Ahhh great catch, you're totally right, I forgot about the NFS requirement because we do things a little different during development. Please disregard!

PR #516, opened 2024-05-23
enable the new prometheus endpoint for Connect

colearendt:
Add a new .prometheus.enabled flag (defaults true) and a .legacyPrometheus flag (defaults false) to revert to old behavior. A simple `legacyPrometheus=true` should revert back to backwards-compatible behavior, but we want folks to switch to the new endpoint, so that is the new default!

Also removed the default / graphite values from the `values.yaml` file for simplicity. As per usual, settings that a user sets in our `.Values.config` will take precedence over convenience helpers like `prometheus.port`

### DIFF
--- a/charts/rstudio-connect/NEWS.md
+++ b/charts/rstudio-connect/NEWS.md
@@ -6,8 +6,8 @@
   - We are now using an internal prometheus endpoint with all new metrics
   - As a result, the `graphiteExporter` sidecar has been removed
   - Some metrics from the `graphiteExporter` will no longer be present
-  - There is also a new "off-switch" for prometheus at `prometheus.enabled=true`
-  - To revert to old behavior, set `legacyPrometheus=true` (and please reach out to let us know why!)
+  - The parent / main "off-switch" for prometheus is at `prometheus.enabled`
+  - To revert to the old exporter, set `prometheus.legacy=true` (and please reach out to let us know why!)
 
 ## 0.6.7
 
--- a/charts/rstudio-connect/NEWS.md
+++ b/charts/rstudio-connect/NEWS.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.7.0
+
+- BREAKING: the prometheus endpoint has changed from port `9108` to `3232` by default
+  - We are now using an internal prometheus endpoint with all new metrics
+  - As a result, the `graphiteExporter` sidecar has been removed
+  - Some metrics from the `graphiteExporter` will no longer be present
+  - There is also a new "off-switch" for prometheus at `prometheus.enabled=true`
+  - To revert to old behavior, set `legacyPrometheus=true` (and please reach out to let us know why!)
+
 ## 0.6.7
 
 - Documentation site updates
--- a/charts/rstudio-connect/README.md
+++ b/charts/rstudio-connect/README.md
@@ -1,6 +1,6 @@
 # Posit Connect
 
-![Version: 0.6.7](https://img.shields.io/badge/Version-0.6.7-informational?style=flat-square) ![AppVersion: 2024.04.1](https://img.shields.io/badge/AppVersion-2024.04.1-informational?style=flat-square)
+![Version: 0.7.0](https://img.shields.io/badge/Version-0.7.0-informational?style=flat-square) ![AppVersion: 2024.04.1](https://img.shields.io/badge/AppVersion-2024.04.1-informational?style=flat-square)
 
 #### _Official Helm chart for Posit Connect_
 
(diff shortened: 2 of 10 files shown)


charts/rstudio-connect/NEWS.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.7.0
+
+- BREAKING: the prometheus endpoint has changed from port `9108` to `3232` by default
+  - We are now using an internal prometheus endpoint with all new metrics
+  - As a result, the `graphiteExporter` sidecar has been removed
+  - Some metrics from the `graphiteExporter` will no longer be present
+  - There is also a new "off-switch" for prometheus at `prometheus.enabled=true`
```
jspiewak, line 9:
```suggestion
  - There is also a new "off-switch" for prometheus at `prometheus.enabled`
```

charts/rstudio-connect/NEWS.md
```diff
@@ -1,5 +1,14 @@
...
+  - To revert to old behavior, set `legacyPrometheus=true` (and please reach out to let us know why!)
```
jspiewak, line 10: If you set this, do you also need to turn set `promethus.enabled=false`? Be more clear

charts/rstudio-connect/README.md
```diff
@@ -142,6 +142,7 @@ The Helm `config` values are converted into the `rstudio-connect.gcfg` service c
 | launcher.templateValues | object | `{"job":{"annotations":{},"labels":{}},"pod":{"affinity":{},"annotations":{},"command":[],"containerSecurityContext":{},"defaultSecurityContext":{},"env":[],"extraContainers":[],"imagePullPolicy":"","imagePullSecrets":[],"initContainers":[],"labels":{},"nodeSelector":{},"priorityClassName":"","securityContext":{},"serviceAccountName":"","tolerations":[],"volumeMounts":[],"volumes":[]},"service":{"annotations":{},"labels":{},"type":"ClusterIP"}}` | Values to pass along to the Posit Connect session templating process |
 | launcher.templateValues.pod.command | list | `[]` | command for all pods. This is really not something we should expose and will be removed once we have a better option |
 | launcher.useTemplates | bool | `true` | Whether to use launcher templates when launching sessions. Defaults to true |
+| legacyPrometheus | bool | `false` | Whether to enable the legacy prometheusExporter INSTEAD OF the built-in product exporter. Takes precedence over    prometheusExporter.enabled |
```
jspiewak, line 145: So this is more clear, but too many spaces.

```suggestion
| legacyPrometheus | bool | `false` | Whether to enable the legacy prometheusExporter INSTEAD OF the built-in product exporter. Takes precedence over prometheusExporter.enabled |
```

charts/rstudio-connect/README.md
```diff
@@ -166,6 +167,7 @@ The Helm `config` values are converted into the `rstudio-connect.gcfg` service c
 | pod.volumes | list | `[]` | An array of maps that is injected as-is into the "volumes:" component of the pod spec |
 | podDisruptionBudget | object | `{}` | Pod disruption budget |
 | priorityClassName | string | `""` | The pod's priorityClassName |
+| prometheus | object | `{"enabled":true,"port":3232}` | Whether to enable prometheus metrics in the product |
 | prometheusExporter.enabled | bool | `true` | Whether the  prometheus exporter sidecar should be enabled |
```
jspiewak, line 171: Should this now default to `false`?

charts/rstudio-connect/templates/deployment.yaml
```diff
@@ -118,6 +124,11 @@ spec:
         imagePullPolicy: "{{ .Values.image.imagePullPolicy }}"
         ports:
         - containerPort: {{ .Values.pod.port }}
+          name: http
+        {{- if and .Values.prometheus.enabled (not .Values.legacyPrometheus) }}
+        - containerPort: {{ .Values.prometheus.port }}
+          name: metrics
```
jspiewak, line 130: Interestingly, neither Chronicle agent nor Datadog needed this for me

charts/rstudio-connect/templates/service-monitor.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.prometheusExporter.enabled .Values.serviceMonitor.enabled -}}
+{{- if and .Values.prometheus.enabled .Values.serviceMonitor.enabled -}}
```
jspiewak, line 1: How does one get legacy behavior here?

charts/rstudio-connect/templates/svc.yaml
```diff
@@ -28,8 +28,7 @@ spec:
     nodePort: {{ .Values.service.nodePort }}
 {{- end }}
     targetPort: {{ .Values.service.targetPort }}
-  {{- if .Values.prometheusExporter.enabled }}
+  {{- if .Values.prometheus.enabled }}
```
jspiewak, line 31: How does one get legacy behavior here?

jspiewak (commented):
Perhaps `prometheusExporter.enabled` and `config.Metrics.GraphiteEnabled` should get the value of `legacyPrometheus`?

jspiewak (approved):
🤞🏻

tnederlof:
@jspiewak and @colearendt should we do something similar for Workbench?

jspiewak:
@tnederlof that would be nice

PR #529, opened 2024-07-10
Update templates

colearendt:
Launcher now ships template versions 2.4.0. The helm charts are now updated accordingly

### DIFF
--- a/charts/rstudio-connect/NEWS.md
+++ b/charts/rstudio-connect/NEWS.md
@@ -4,7 +4,8 @@
 
 - Bump version of launcher templates `job.tpl` and `service.tpl`
   - `enableServiceLinks` now defaults to `false` for sessions (instead of not being set).
-    To enable them for sessions, set `launcher.templateValues.enableServiceLinks: true`
+    To enable them for sessions, set `launcher.templateValues.enableServiceLinks: true`.
+  - Also see related discussion [on the Kubernetes project](https://github.com/kubernetes/kubernetes/issues/121787)
 
 ## 0.7.4
 
--- a/charts/rstudio-connect/NEWS.md
+++ b/charts/rstudio-connect/NEWS.md
@@ -1,3 +1,7 @@
+## 0.7.4
+
+- Bump version of launcher templates `job.tpl` and `service.tpl`
+
 # 0.7.3
 
 - Bump Connect version to 2024.06.0
--- a/charts/rstudio-connect/NEWS.md
+++ b/charts/rstudio-connect/NEWS.md
@@ -1,12 +1,14 @@
+# Changelog
+
 ## 0.7.4
 
 - Bump version of launcher templates `job.tpl` and `service.tpl`
+  - `enableServiceLinks` now defaults to `false` for sessions (instead of not being set).
+    To enable them for sessions, set `launcher.templateValues.enableServiceLinks: true`
 
-# 0.7.3
+## 0.7.3
 
 - Bump Connect version to 2024.06.0
-
-# Changelog
 
 ## 0.7.2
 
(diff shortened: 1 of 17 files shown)


dbkegley:
Thank you @colearendt ! I started to do this like 3 separate times but got lost in the sauce on the template merge each time. If you have a minute, could you write up a summary of the commands you use somewhere to handle the diff?

dbkegley (approved):

charts/rstudio-connect/NEWS.md
```diff
@@ -1,3 +1,7 @@
+## 0.7.4
```
jspiewak, line 1: This (and 0.7.3) should go after `# Changelog`, no?
and 0.7.3 should use double-hash for h2 heading.

dbkegley: Good catch. We missed [this change](https://github.com/rstudio/helm/commit/8a327f31196c235b08f08f57996d56ae9434acff) which added the changelog header. I'll update the Connect release automation to account for this when we write the news entry.

dbkegley: @colearendt I've updated our automation in connect so that this wont happen during our next release but it does depend on this PR getting merged because `# Changelog` needs to be at the beginning of the NEWS file.

charts/rstudio-workbench/files/job.tpl
```diff
@@ -77,6 +83,7 @@ spec:
       {{- if .Job.host }}
       nodeName: {{ toYaml .Job.host }}
       {{- end }}
+      enableServiceLinks: {{ if hasKey $templateData.pod "enableServiceLinks" }}{{ $templateData.pod.enableServiceLinks }}{{ else }}false{{ end }}
```
colearendt, line 86: @jspiewak What do you think about this?

"Proper" backwards compatibility would mean leaving this unset if it is not specified, but this is a better security posture so you do not expose the endpoints of your other services on Kubernetes to arbitrary Connect content.

dbkegley: I can't speak for workbench but for connect I think it's fair to say that these content pods are an implementation detail and shouldn't be considered a part of our public API. I'm fine with changing the default in Connect

dbkegley (approved):

charts/rstudio-workbench/files/job.tpl
jspiewak: What is the behavior when unset?
What would a customer experience with this unexpectedly defaulting to `false`?

jspiewak: So, you could roll this out in two steps:
1. Enable control over the setting, but maintain existing behavior
2. Move to defaulting to `false`

Two steps is more work, but gives consumers additional control.
Not clear to me if it is worth it. What is the likelihood that anyone is actually relying upon it being true?

Yes, very interesting discussion. Might be worth linking to in the NEWS item you have.

colearendt: I think it's extremely unliklely that anyone is relying upon it being true. And I like the idea of just flipping the switch so (1) maybe we get some data if they are, (2) we ensure everyone has a "more secure" / more stable environment.

I'll definitely add to NEWS!

jspiewak:
@jforest do you want to get this landed soon while it is relatively fresh? Was it something you covered with Cole?

jforest:
> @jforest do you want to get this landed soon while it is relatively fresh? Was it something you covered with Cole?

It was something we covered, and added his latest commits while doing so.  It appears there are new conflicts we need to fix first, I'll look at that now

jspiewak (approved):

Issue #490, opened 2024-05-02
Add new Workbench prometheus endpoint


tnederlof:
Workbench 2024.04.0 released a new Prometheus endpoint as a preview feature: https://docs.posit.co/ide/server-pro/auditing_and_monitoring/prometheus_metrics.html#prometheus-metrics

This would be especially helpful to have in the helm charts as many using Workbench in K8s use Prometheus for metrics. Since this is preview probably still worth keeping the graphite exporter as the default but allow an easy way to opt into the native Prometheus endpoint instead.

PR #532, opened 2024-07-11
Enable the new prometheus endpoint for Workbench

tnederlof:
This PR follows closely what @colearendt did for Connect here: https://github.com/rstudio/helm/pull/516
Closes: https://github.com/rstudio/helm/issues/490

Add a new .prometheus.enabled flag (defaults true) and a .legacyPrometheus flag (defaults false) to revert to old behavior. A simple `legacyPrometheus=true` should revert back to backwards-compatible behavior, but we want folks to switch to the new endpoint, so that is the new default!

Also removed the default / graphite values from the `values.yaml` file for simplicity. As per usual, settings that a user sets in our `.Values.config` will take precedence over convenience helpers like `prometheus.port`

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-workbench
 description: Official Helm chart for Posit Workbench
-version: 0.7.6
+version: 0.8.0
 apiVersion: v2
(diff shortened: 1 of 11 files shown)


colearendt (approved):
BOOM! Looks awesome! Well done! 🎉

Issue #520, opened 2024-05-31
Allow sssd.conf ldap_default_authtok to reference a kubernetes secret


samcofer:
Currently, this password, whether encrypted or unencrypted must be passed as a raw value in the userProvisioning section of the chart, which limits the transportability of this chart and prevents/restricts the ability for administrators to check their values.yaml into version control.

PR #535, opened 2024-07-19
PWB: Support defining sensitive config values via k8s secrets

pat-s:
Applies to

- `launcherPem`
- `secureCookieKey`
- `userPassword`
- `database.conf`

When doing so, the `rstudio-secret` is skipped and individual mounts containing the above-mentioned secret values are injected.

This PR is breaking as the items listed are now maps and existing deployments would need to migrate from `launcherPem: <value>` to `launcherPem.value: <value>`.

I've tested this in a deployment of mine and it works as expected so far.

```yaml
secureCookieKey:
  existingSecret: secure-cookie-key
launcherPem:
  existingSecret: launcher-pem
config:
  database:
    conf:
      existingSecret: database-conf
userPassword:
  existingSecret: workbench-user-password
```

The advantage of this approach is that it gives power to the users and makes use of "simple" k8s secrets instead of relying on some chart magic which puts together a bundled secret from plain config values.

related #520 #493

### DIFF
--- a/charts/rstudio-workbench/README.md
+++ b/charts/rstudio-workbench/README.md
@@ -14,7 +14,6 @@ To ensure a stable production deployment:
 * "Pin" the version of the Helm chart that you are using. You can do this using the:
   * `helm dependency` command *and* the associated "Chart.lock" files *or*
   * the `--version` flag.
-
     ::: {.callout-important}
     This protects you from breaking changes.
     :::
--- a/charts/rstudio-workbench/README.md
+++ b/charts/rstudio-workbench/README.md
@@ -14,7 +14,7 @@ To ensure a stable production deployment:
 * "Pin" the version of the Helm chart that you are using. You can do this using the:
   * `helm dependency` command *and* the associated "Chart.lock" files *or*
   * the `--version` flag.
- 
+
     ::: {.callout-important}
     This protects you from breaking changes.
     :::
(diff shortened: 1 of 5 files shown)


charts/rstudio-workbench/README.md
```diff
@@ -507,7 +509,8 @@ Use of [Sealed secrets](https://github.com/bitnami-labs/sealed-secrets) disables
...
 | launcher.templateValues.pod.command | list | `[]` | command for all pods. This is really not something we should expose and will be removed once we have a better option |
 | launcher.useTemplates | bool | `false` | whether to render and use templates in the job launching process |
-| launcherPem | string | `""` | An inline launcher.pem key. If not provided, one will be auto-generated. See README for more details. |
+| launcherPem.value | string | `""` | An inline launcher.pem key. If not provided, one will be auto-generated. See README for more details. |
+| launcherPem.existingSecret | string | `""` | Existing Secret for launcherPem |
```
GCRev, line 513: Is this like the others in that it's a secret containing the contents of `launcherPem`? It might be worth clarifying that in the phrasing of the summary of the setting.
```suggestion
| launcherPem.existingSecret | string | `""` | Existing secret containing launcherPem contents |
```

damienleger-cure51:
Hello, I would love to see this merged! 🙏 

On my setup, helm chart is deployed through ArgoCD and it [doesn't support helm lookup function](https://github.com/argoproj/argo-cd/issues/5202). This lookup function is the one responsible for checking that launcher.pem+secureCookieKey exists, resulting of those secret values regenerated on every ArgoCD sync. Having those to be set only directly on values.yaml currently to fix this situation is a problem for sensitive value.

samcofer:
@jforest Can you take another look at this PR?

jforest (changes requested):
@pat-s You also need to update `NEWS.md`

We need to draw attention to the breaking changes.  Things like `global.secureCookieKey` is no longer a string, and that string has moved to `global.secureCookieKey.value`

pat-s:
> This project is still pre-1.0.0 so technically considered under "initial development". See [semver.org#spec-item-4](https://semver.org/#spec-item-4):

To me this general argumentation doesn't make sense as it just undermines all ideas of semver when doing so over many years - especially when a project/repo is in "active" use for many years already. Even more if its an enterprise product that should not be in "initial development" ever arguably. But I am not a maintainer, just a person sharing his thoughts.

(not meant offensive against you @fh-albert-lee, just took your comment as informative)

---

Anyhow, hope we can get this merged soon. Maintainers: feel free to adjust versions and changelog/README entries to your liking, this doesn't need to come from my side and there's no value in me solving Chart.yaml conflicts repeatedly.

PR #558, opened 2024-09-26
Fix: Set the `securityContext.fsGroup` to default `999` guid

tylfin:
Bumps the PPM chart to `0.5.34`

Refs: https://github.com/rstudio/package-manager/issues/14422

### DIFF
--- a/charts/rstudio-pm/values.yaml
+++ b/charts/rstudio-pm/values.yaml
@@ -129,6 +129,7 @@ pod:
   # -- the [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for the main Package Manager container. Evaluated as a template.
   containerSecurityContext:
     runAsUser: 999
+    fsGroup: 999
     runAsNonRoot: true
     allowPrivilegeEscalation: false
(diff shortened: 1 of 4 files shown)


CDRayn (approved):

charts/rstudio-pm/values.yaml
```diff
@@ -129,6 +129,7 @@ pod:
   # -- the [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for the main Package Manager container. Evaluated as a template.
   containerSecurityContext:
     runAsUser: 999
+    fsGroup: 999
```
glin, line 132: @tylfin I updated to the newest chart and got a warning about an unknown field:
```sh
W0927 16:29:42.405025  521699 warnings.go:70] unknown field "spec.template.spec.containers[0].securityContext.fsGroup"
```

I think `fsGroup` has to go in the `securityContext` block above, not `containerSecurityContext`?

tylfin: Ah okay, let me patch this

PR #567, opened 2024-10-07
Alert on GHA failures for the chart testing workflow

jforest:
Send alerts to the posit-helm channel when linting or installation GHA fail


jforest:
Sorry @rstudio/ptd-dev I had put the incorrect team in the CODEOWNERS before, you do not have to do anything!

PR #570, opened 2024-10-08
Fix: PPM README should document lowercase postgres

tylfin:
Refs: https://github.com/rstudio/package-manager/issues/14420#issuecomment-2398208531


jmwoliver:
Oh I see this PR now: https://github.com/rstudio/package-manager/pull/14426

PR #574, opened 2024-10-09
Add notifications for failing gha for all workflows

jforest:
Very basic, just pointing to the logs for each failed run.


jspiewak (approved):
Depending on how chatty the alerts end up getting, we may want to have a `#helm-bots` channel.

PR #581, opened 2024-10-11
Re-enable chart install tests, but exclude all currently failing charts

jforest:
Marking this as ready to review for feedback on the idea.

The current set of values files in the `ci` subdirectory of the charts are fine for linting, but many values are nonsense and end up causing installs to fail. 

My idea is that we make an entirely new set of values files that are just for installs (they should pass linting anyway right?) while keeping the existing values files.  The one change will be to rename all the current values files to `lint-<existingname>.yaml` so they get deleted in a step of the install workflow before the `ct install` tests run


.github/workflows/chart-test.yaml
```diff
@@ -71,82 +73,90 @@ jobs:
...
+        if: ${{ steps.list-changed.outputs.changed == 'true' || github.ref == 'refs/heads/main' }}
+        run: |
+          helm repo add sealed-secrets https://bitnami-labs.github.io/sealed-secrets
+          helm install sealed-secrets sealed-secrets/sealed-secrets
+
+      # Delete any values files that are viable for linting only.  That way we only have the values files that are used
+      # for installation.
+      - name: Remove linting-only ci values files
+        if: ${{ steps.list-changed.outputs.changed == 'true' || github.ref == 'refs/heads/main' }}
+        run: rm charts/*/ci/lint-*
```
jspiewak, line 120: I don't love deleting files that are part of the checkout.
I would rather see if we can have separate directories for values files used for linting and use CLI arguments to select the right set of things for each type of test.

jforest: So far I haven't been able to find a way to configure the tool to use anything other than the `ci` directory.  I will keep looking

jforest:
After talking with @jspiewak I'm going to close this PR, and open a new one where I will move the current `ci` directory to `ci-lint` and create a new `ci-install` directory within which I will create install test values files.  

Then I will need to modify the workflows to symlink the proper one in place before running lints or installs.  (and clean up after themselves when done).  Also will need to adjust the Justfile to do the right thing.

PR #591, opened 2024-10-17
Test the license-file-secret-values linting file with basic-values.yaml additions

jforest:
Since we no longer have the `ci` directories in each chart directory, we need to run the install all tests every time.  In order to do that we need the kind cluster to be up, and if we're doing that we might as well install the sealed secrets helm chart too.


.github/workflows/chart-test.yaml
```diff
@@ -130,26 +130,38 @@ jobs:
...
 
+      - name: Create posit-test namespace
+        run: kubectl create namespace posit-test
+
+      - name: Create License File Secrets
+        run: |
+          echo "${{ secrets.PWB_LICENSE_FILE }}" > pwb.lic
+          kubectl create secret generic pwb-license --from-file=pwb.lic --namespace posit-test
+          rm pwb.lic
+          echo "${{ secrets.RSC_LICENSE_FILE }}" > rsc.lic
```
jspiewak, line 147: I think they refer to it as PCT (Posit Connect) now

jspiewak (approved):

.github/workflows/chart-test.yaml
jforest: Thanks!

aronatkins: 😭 We are the Posit Percent? 🤮 @edavidaja 

It might still be "RSC", as hinted by this reference:
https://github.com/rstudio/helm/blob/b578ab99908643f3c6aa0d91b4b04fc0cb7d83e7/charts/rstudio-connect/templates/deployment.yaml#L107

jforest: The license file I generated from https://connect.posit.it/support/license_management/ is named with `PCT` in it 

image

PPM has PPM, and PWB has PWB.  🤷

aronatkins: Yeah. PCT is being adopted by folks. The `envVarPrefix` still uses "RSC".

PR #592, opened 2024-10-18
Fix: Use `fsGroupChangePolicy` to prevent recursive perm changes

tylfin:
See https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods for more information

Refs: https://github.com/rstudio/package-manager/issues/14547


glin (approved):
Tested and confirmed that this fixed the issue when repro-ing on Azure K8s with an Azure Disks volume. Also tested this with Azure Files NFS volumes, but the issue never happened with NFS volumes.

tylfin:
@jforest What do I need to do to get the linter to pass here?

PR #593, opened 2024-10-22
Fix connect pod service account values when launcher is enabled

### LINKED ISSUE #584: [CONNECT] Service account name is not properly set
The chart ignores the service account name provided in chart values and is always set to the result of `rstudio-connect.fullname` template.

The following line seems to be incorrect: https://github.com/rstudio/helm/blob/f7bf50ca6a29a455122e75edf47fd5102fddb16b/charts/rstudio-connect/templates/rbac.yaml#L4 

Based on the [Helm documentation](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) for `default` function, its syntax is `default DEFAULT_VALUE GIVEN_VALUE`.

Therefore to take properly into account the service account name specified in Values, the line above should rather be:
```
{{ $serviceAccountName := default (include "rstudio-connect.fullname" .) .Values.rbac.serviceAccount.name }}
```

Similar issue is present here:
https://github.com/rstudio/helm/blob/f7bf50ca6a29a455122e75edf47fd5102fddb16b/charts/rstudio-connect/templates/deployment.yaml#L80

dbkegley:
Fixes #584 

The connect pod service account values were being incorrectly applied when the job launcher is enabled. This PR fixes the Connect pod's serviceAccount value and removes some dead code.

Before:

```shell
# w/o launcher
» helm template ./ --set rbac.serviceAccount.name=asdf | grep serviceAccount 
      serviceAccountName: "asdf"

# w/ launcher
» helm template ./ --set launcher.enabled=true --set sharedStorage.create=true --set rbac.serviceAccount.name=asdf | grep serviceAccount
   ...
      serviceAccountName: release-name-rstudio-connect
```

After:

```shell
# w/o launcher
» helm template ./ --set rbac.serviceAccount.name=asdf | grep serviceAccount 
      serviceAccountName: "asdf"

# w/ launcher
» helm template ./ --set launcher.enabled=true --set sharedStorage.create=true --set rbac.serviceAccount.name=asdf | grep serviceAccount
...
      serviceAccountName: asdf
```

### DIFF
--- a/charts/rstudio-connect/templates/deployment.yaml
+++ b/charts/rstudio-connect/templates/deployment.yaml
@@ -77,7 +77,7 @@ spec:
          * https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-multiple-service-accounts
          */}}
       {{- if and .Values.rbac.create .Values.launcher.enabled }}
-      {{ $serviceAccountName := default (default .Values.rbac.serviceAccount.name .Values.pod.serviceAccountName) (include "rstudio-connect.fullname" .) }}
+      {{ $serviceAccountName := default (include "rstudio-connect.fullname" .) .Values.rbac.serviceAccount.name }}
       serviceAccountName: {{ $serviceAccountName }}
       {{- else }}
       serviceAccountName: {{ .Values.rbac.serviceAccount.name | toString | quote }}
--- a/charts/rstudio-connect/templates/rbac.yaml
+++ b/charts/rstudio-connect/templates/rbac.yaml
@@ -1,7 +1,7 @@
 {{- if and (.Values.rbac.create) (.Values.launcher.enabled) }}
 {{ $namespace := $.Release.Namespace }}
 {{ $targetNamespace := default $.Release.Namespace .Values.launcher.namespace }}
-{{ $serviceAccountName := default .Values.rbac.serviceAccount.name (include "rstudio-connect.fullname" .) }}
+{{ $serviceAccountName := default (include "rstudio-connect.fullname" .) .Values.rbac.serviceAccount.name }}
 {{ $serviceAccountCreate := .Values.rbac.serviceAccount.create }}
 {{ $serviceAccountAnnotations := .Values.rbac.serviceAccount.annotations }}
 {{ $serviceAccountLabels := .Values.rbac.serviceAccount.labels }}
--- a/.github/workflows/chart-test.yaml
+++ b/.github/workflows/chart-test.yaml
@@ -46,7 +46,12 @@ jobs:
       - name: Run chart-testing (lint changed)
         id: ct-lint
         if: steps.list-changed.outputs.changed == 'true'
-        run: ct lint --target-branch main --chart-dirs charts --chart-dirs other-charts
+        run: |
+          {
+            echo 'result<<EOF'
+            ct lint --target-branch main --chart-dirs charts --chart-dirs other-charts
+            echo 'EOF'
+          } >> $GITHUB_OUTPUT
         continue-on-error: true
 
       - name: Run chart-testing (lint all)
--- a/.github/workflows/chart-test.yaml
+++ b/.github/workflows/chart-test.yaml
@@ -46,12 +46,7 @@ jobs:
       - name: Run chart-testing (lint changed)
         id: ct-lint
         if: steps.list-changed.outputs.changed == 'true'
-        run: |
-          {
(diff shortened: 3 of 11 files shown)


charts/rstudio-connect/templates/deployment.yaml
```diff
@@ -77,7 +77,7 @@ spec:
          * https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-multiple-service-accounts
          */}}
       {{- if and .Values.rbac.create .Values.launcher.enabled }}
-      {{ $serviceAccountName := default (default .Values.rbac.serviceAccount.name .Values.pod.serviceAccountName) (include "rstudio-connect.fullname" .) }}
```
dbkegley, line 80: `pod.serviceAccountName` is already fully deprecated

```
» helm template ./ --set pod.serviceAccountName=asdf                  
Error: execution error at (rstudio-connect/templates/NOTES.txt:55:4): 

`pod.serviceAccountName` is no longer used. Use `rbac.serviceAccount.name` instead.

Use --debug flag to render out invalid YAML
```

zackverham:
Not immediately related to your PR itself - but it does feel like something in CI should have caught this, yeah?

dbkegley:
@jforest I'm not very familiar with `chart-testing` but do you know if it would allow us to provide assertions for expected output?  This bug for example would have produced a "valid" installation but would have used the wrong service account for the Connect pod.  I'm happy to help look into this more myself as well, just wondering if you know off hand

jforest:
> @jforest I'm not very familiar with `chart-testing` but do you know if it would allow us to provide assertions for expected output? This bug for example would have produced a "valid" installation but would have used the wrong service account for the Connect pod. I'm happy to help look into this more myself as well, just wondering if you know off hand

I don't think so, I believe you can create a values file (ending in `-values.yaml`) and run an install with those values files. 

Could you merge main into this branch please?  It should fix the linting error

dbkegley:
> I don't think so, I believe you can create a values file (ending in -values.yaml) and run an install with those values files.

That makes sense, seems like we have the install half but are missing the assertions.  I believe we just need to write some tests that live in the `charts/rstudio-connect/templates/test` directory and it sounds like they will get evaluated automatically by each `ct install`

https://helm.sh/docs/topics/chart_tests/
https://github.com/helm/chart-testing/blob/main/doc/ct_install.md#synopsis

cc @plascaray 

> Could you merge main into this branch please? It should fix the linting error

Done!

dbkegley:
@jforest would you mind taking another look at this CI failure? I pulled from main and the branch seems to have passed all the linting but it's failing to add the linting results to the PR with the following error:

`SyntaxError: Unexpected identifier 'pod'`

jforest:
> @jforest would you mind taking another look at this CI failure? I pulled from main and the branch seems to have passed all the linting but it's failing to add the linting results to the PR with the following error:
> 
> `SyntaxError: Unexpected identifier 'pod'`

Yah, I've got a test PR trying to fix this, I think I'm just going to pull out trying to post the output to the PR, it's causing more trouble than it's worth.  People can just check the GHA output directly

jforest:
@dbkegley I have merged https://github.com/rstudio/helm/pull/597, you should merge `main` into your branch!

dbkegley:
Thanks @jforest ! It's working now, would you mind reviewing/approving for infraops so we can get this fix released?

charts/rstudio-connect/templates/rbac.yaml
```diff
@@ -1,7 +1,7 @@
 {{- if and (.Values.rbac.create) (.Values.launcher.enabled) }}
 {{ $namespace := $.Release.Namespace }}
 {{ $targetNamespace := default $.Release.Namespace .Values.launcher.namespace }}
-{{ $serviceAccountName := default .Values.rbac.serviceAccount.name (include "rstudio-connect.fullname" .) }}
+{{ $serviceAccountName := default (include "rstudio-connect.fullname" .) .Values.rbac.serviceAccount.name }}
```
jforest, line 4: The changes of order here are just to ensure that if we have `.Values.rbac.serviceAccount.name` set, it will override the default of `rstudio-connect.fullname`?

jforest (approved):

charts/rstudio-connect/templates/rbac.yaml
dbkegley: Yep, that's exactly right

PR #599, opened 2024-10-30
Add helm unittesting scaffolds

dbkegley:
related to https://github.com/rstudio/helm/pull/593

Our helm charts are missing tests that can assert that the rendered helm chart templates produce the correct output based on the inputs. This PR sets up unittest scaffolds for all product chart owners to start adding these types of tests

We use the [helm unittest plugin](https://github.com/rstudio/helm/pull/593) for writing tests. Tests live in the `charts/*/test` directory for each helm chart.


.github/workflows/chart-test.yaml
```diff
@@ -79,6 +79,29 @@ jobs:
...
+      - name: Install helm unittest plugin
+        run: helm plugin install --version v0.6.3  https://github.com/helm-unittest/helm-unittest.git
+
+      - name: Run chart unit tests
+        run: |
+          for dir in $(ls -d charts/*/); do
+            pushd $dir; helm dependencies update; popd
+            helm unittest $dir
+          done
+
```
jforest, line 104: We need to add some alerting to the tests , similar to what we do for the install and lint tests: https://github.com/rstudio/helm/blob/884427b511575bd2bf7ae1149e7db1106f12b8c9/.github/workflows/chart-test.yaml#L154-L173

jforest (changes requested):
Need to bump the chart version

```
✖︎ rstudio-connect => (version: "0.7.12", path: "charts/rstudio-connect") > chart version not ok. Needs a version bump! 
```

jspiewak:
> Need to bump the chart version
> 
> ```
> ✖︎ rstudio-connect => (version: "0.7.12", path: "charts/rstudio-connect") > chart version not ok. Needs a version bump! 
> ```

Perhaps if the `tests` directory is added to `.helmignore` per https://github.com/helm-unittest/helm-unittest?tab=readme-ov-file#get-started then we won't have to bump the chart version when adding tests (which feels like a bad thing)?

dbkegley:
> Perhaps if the tests directory is added to .helmignore

Looks like changing the `.helmignore` means we have to bump the version initially, which I guess makes sense because it could impact packaging. I'm going to bump Connect's version here and I'll leave it to other chart owners to decide when they want to make this change in their charts to add tests.

jspiewak:
> > Perhaps if the tests directory is added to .helmignore
> 
> Looks like changing the `.helmignore` means we have to bump the version initially, which I guess makes sense because it could impact packaging. I'm going to bump Connect's version here and I'll leave it to other chart owners to decide when they want to make this change in their charts to add tests.

Yeah, makes sense that the addition of `.helmignore` requires a version bump. I wonder if a change to `.helmignore` would require and additional bump, my guess is yes. But that changes to anything it is ignoring would not.

jforest:
> Looks like changing the `.helmignore` means we have to bump the version initially, which I guess makes sense because it could impact packaging. I'm going to bump Connect's version here and I'll leave it to other chart owners to decide when they want to make this change in their charts to add tests.

I don't see the downside in adding `.helmignore` for each chart now, to make it simpler for everyone else to get started in the future.

Yes, it's a version bump right now which is slightly annoying.  But it makes everything easier for those coming after you, they won't need to know to put the `.helmignore` file in place.  They can just get busy writing tests!

dbkegley:
> > Looks like changing the `.helmignore` means we have to bump the version initially, which I guess makes sense because it could impact packaging. I'm going to bump Connect's version here and I'll leave it to other chart owners to decide when they want to make this change in their charts to add tests.
> 
> I don't see the downside in adding `.helmignore` for each chart now, to make it simpler for everyone else to get started in the future.
> 
> Yes, it's a version bump right now which is slightly annoying. But it makes everything easier for those coming after you, they won't need to know to put the `.helmignore` file in place. They can just get busy writing tests!

Works for me!

CONTRIBUTING.md
```diff
@@ -34,6 +34,19 @@ We'll try to be as responsive as possible in reviewing and accepting pull reques
...
 
+## Testing
+
+Running the [helm chart unit tests](https://github.com/helm-unittest/helm-unittest):
+
+```
+# install the unittest plugin
+helm plugin install https://github.com/helm-unittest/helm-unittest.git
+
+# run the rstudio-connect chart unittests
```
jforest, line 45: Maybe make this a little more generic so people know it's not just for `rstudio-connect`?

dbkegley: Added a `just test` target to our justfile

bschwedler (approved):
Thank you for putting this together!

So much cleaner and targeted than having an ever-growing collection of values files & golden render output for testing.

jforest (approved):
Thank you @dbkegley!  This is a fantastic addition that will make it easier for all the teams to start unit testing their charts, I really appreciate it!

PR #615, opened 2024-11-15
Add a few more unittests for rstudio-workbench

jforest:


GCRev (approved):
LGTM. I'm assuming that the Legacy Port values are not configurable and that they're always going to 9108.

bschwedler (approved):

PR #624, opened 2024-12-02
Add yaml & quarto markdown in GH statistics

Cecilsingh:
Adding `.yaml` and `.qmd` file extensions into `.gitattributes` so that it's recognized in github statistics.


jforest (approved):
The install test failures and doc failures are expected when a PR is coming from a forked repo.  This looks good!

PR #625, opened 2024-12-04
Adding more unit tests for the rstudio-workbench chart

jforest:
I found what I consider to be a bug: https://github.com/rstudio/helm/blob/a34cf7f18525cca9b70cdffff723508f691b2185/charts/rstudio-workbench/templates/_helpers.tpl#L186-L187

I've added that fix here, and all the existing unit tests continue to work so, huzzah!  Tests doing their thing

Otherwise, added more tests to attempt to cover all of the branching in `rstudio-workbench.containers` in `_helpers.tpl`

### DIFF
--- a/charts/rstudio-workbench/templates/_helpers.tpl
+++ b/charts/rstudio-workbench/templates/_helpers.tpl
@@ -183,8 +183,8 @@ containers:
       cpu: "{{ .Values.resources.requests.cpu }}"
       ephemeral-storage: "{{ .Values.resources.requests.ephemeralStorage }}"
     {{- end }}
-    limits:
     {{- if .Values.resources.limits.enabled }}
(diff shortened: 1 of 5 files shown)


charts/rstudio-workbench/templates/_helpers.tpl
```diff
@@ -183,8 +183,8 @@ containers:
       cpu: "{{ .Values.resources.requests.cpu }}"
       ephemeral-storage: "{{ .Values.resources.requests.ephemeralStorage }}"
     {{- end }}
-    limits:
     {{- if .Values.resources.limits.enabled }}
+    limits:
```
GCRev, line 187: 👀

GCRev (approved):
LGTM

PR #629, opened 2024-12-16
Bump launcher templates for Workbench chart to 2.5.0

zachhannum:
Adds init containers support from the Job spec from rstudio/launcher#1070

This should go out with the other 2024.12.0 post-release changes.

PR #631, opened 2024-12-17
Bump pwb 2024.12.0

melissa-barca:
- bump pwb to 2024.12.0
- update vs code default config values to match those in workbench
- add positron config files commented out in case customers want to try it out while the feature is in preview
- pull in changes from https://github.com/rstudio/helm/pull/629

### DIFF
--- a/charts/rstudio-workbench/NEWS.md
+++ b/charts/rstudio-workbench/NEWS.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.8.10
+
+- Bump Workbench version to 2024.12.0
+
 ## 0.8.9
 
 - Fix a logic bug where the resource limit key was set even if `resources.limits.enabled` is false
--- a/charts/rstudio-workbench/NEWS.md
+++ b/charts/rstudio-workbench/NEWS.md
@@ -3,7 +3,9 @@
 ## 0.8.10
 
 - Bump Workbench version to 2024.12.0
-
+- Bump version of launcher templates `job.tpl` and `service.tpl`
+  - Populate pod `initContainers` from `.Job.initContainers`
+ 
 ## 0.8.9
 
 - Fix a logic bug where the resource limit key was set even if `resources.limits.enabled` is false
--- a/charts/rstudio-workbench/values.yaml
+++ b/charts/rstudio-workbench/values.yaml
@@ -463,14 +463,28 @@ config:
       default-session-cluster: Kubernetes
     vscode.conf:
       enabled: 1
-      exe: /opt/code-server/bin/code-server
-      args: --host=0.0.0.0
+      session-timeout-kill-hours: 12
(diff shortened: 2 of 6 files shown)


charts/rstudio-workbench/NEWS.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.8.10
+
+- Bump Workbench version to 2024.12.0
+
```
zachhannum, line 6:
```suggestion
- Bump version of launcher templates `job.tpl` and `service.tpl`
  - Populate initContainers from `.Job.initContainers`
```

charts/rstudio-workbench/values.yaml
```diff
@@ -463,14 +463,28 @@ config:
       default-session-cluster: Kubernetes
     vscode.conf:
       enabled: 1
-      exe: /opt/code-server/bin/code-server
-      args: --host=0.0.0.0
+      session-timeout-kill-hours: 12
```
GCRev, line 466: I probably missed some comms on this, but I thought we weren't going to enable this setting by default. I guess it's probably fine to make the timeout period this long by default though.

GCRev (approved):
Seems like were' good to go.

charts/rstudio-workbench/values.yaml
melissa-barca: I thought we agreed with setting it to a large value. @cm421 can you confirm?

cm421: By default it is disabled in Workbench's settings, but for our helm charts we decided to set it to 12. It can always be changed if someone doesn't like it

GCRev: That makes sense!

charts/rstudio-workbench/values.yaml
```diff
@@ -463,14 +463,28 @@ config:
...
+    #    enabled: 1
+    #  positron.extensions.conf: |
+    #    posit.shiny
+    #    posit.publisher
+    #  positron-user-settings.json: |
+    #    {
+    #          "terminal.integrated.shell.linux": "/bin/bash",
+    #          "extensions.autoUpdate": false,
+    #          "extensions.autoCheckUpdates": false
+    #    }
```
cm421, line 487: I'm assuming the intent here is to have Positron disabled by default, but very easy to enable if someone wants it?

cm421 (approved):
~Assuming we like the positron.conf commented out,~ LGTM!

charts/rstudio-workbench/values.yaml
melissa-barca: Correct

cm421: I _just_ was reading your original PR description and saw your bullet point on this. My bad!

PR #632, opened 2024-12-18
Update Workbench Examples

samcofer:
This PR creates a new example that creates user home directories with 700 permissions, which is a common customer request.

It also updates the older auto-scaling example to use templates for that annotation, rather than job-json-overrides

### DIFF
--- a/examples/auto-scaling/rstudio-workbench/values.yaml
+++ b/examples/auto-scaling/rstudio-workbench/values.yaml
@@ -36,8 +36,11 @@ service:
 
 launcher:
   enabled: true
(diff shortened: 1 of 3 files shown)


tnederlof (approved):
Looks good to me! When I did a big revamp of the examples a few months ago I didn't touch the auto-scaling folder so good to update. It would be ideal to work into the doc site https://docs.posit.co/helm/examples/workbench/ but its not under workbench so maybe that's another project to move it under the workbench folder and add a .qmd like the other.

jforest (approved):

PR #648, opened 2025-03-21
PPM: implement service.targetPort as configurable

kmasiello:
service.targetPort was hard coded as port 4242. Modifed this to be a configurable value to support use cases when customer requires routing to alternate port.

### DIFF
--- a/charts/rstudio-pm/values.yaml
+++ b/charts/rstudio-pm/values.yaml
@@ -85,6 +85,8 @@ service:
   nodePort: false
   # -- The Service port. This is the port your service will run under.
   port: 80
+  # -- The port to forward to on the Package Manager pod. Also see pod.port
+  targetPort: 4242
 
 # -- replicas is the number of replica pods to maintain for this service
 replicas: 1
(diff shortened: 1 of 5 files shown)


charts/rstudio-pm/values.yaml
```diff
@@ -85,6 +85,8 @@ service:
   nodePort: false
   # -- The Service port. This is the port your service will run under.
   port: 80
+  # -- The port to forward to on the Package Manager pod. Also see pod.port
```
tnederlof, line 88: There actually isn't a `pod.port` option in the values like in Connect. It looks like `4242` is hard coded here: https://github.com/rstudio/helm/blob/9ee39769c91cea59a705eac4e1578ecab727b26f/charts/rstudio-pm/templates/deployment.yaml#L109

Might make sense to add this value as well.

charts/rstudio-pm/values.yaml
```diff
@@ -85,6 +85,8 @@ service:
...
+  targetPort: 4242
```
glin, line 89: Besides `service.targetPort` and `pod.port`, there are a bunch of other `4242` references in the values, like the `config.HTTP.Listen = :4242` setting, liveness/startup probe. I assume users would have to find-and-replace all `4242` references manually to fully change the running port?

Is it possible to use a template variable here so you only have to set `4242` once somewhere?

Otherwise this is fine of course, as it's still better than what we had before.

glin (approved):

PR #649, opened 2025-03-25
PWB: add `targetPort`, `hostAliases` fix `pod.port`

kmasiello:
`service.targetPort` was hard coded as port 8787. Modified this to be a configurable value to support use cases when customer requires routing to alternate port.

Added `pod.hostAliases` and `launcher.templateValues.pod.hostAliases`

Also resolved issue where `pod.port` was not mapped in `_helpers.tpl`


GCRev (approved):
LGTM

tnederlof (approved):

PR #653, opened 2025-03-31
PWB: Set `launcher.useTemplates: true` by default

tnederlof:
This PR sets the default to `launcher.useTemplates: true` in the Workbench chart. This bumps the chart a minor version to `0.9.0` since it could be breaking if someone is using `job-json-overrides` because both cannot be set simultaneously. There is a [check in the helm chart](https://github.com/rstudio/helm/blob/228a98e729f9e2d0de9c16898f79dfe73759ad01/charts/rstudio-workbench/templates/NOTES.txt#L42) that will error if both `job-json-overrides` and `launcher.useTemplates=true` are defined, this PR adds to that error to note the chart version where this default was changed.

This being `false` by default is the most common root cause of customer issues we see in the field because without it many other chart defaults and examples Posit puts out will subtly not be applied (for example, session changes to `pip.conf`, `imagePullSecrets`, mounting storage to sessions etc.).

Historically before `launcher.templateValues` existed setting `job-json-overrides` was the way to customize jobs and it is still what is documented in our [helm chart README](https://github.com/rstudio/helm/tree/main/charts/rstudio-workbench#job-json-overrides). This PR removes the `job-json-overrides` language from the README and instead instructs users to use `launcher.templateValues` to customize the job templates.

Also worth noting that the [Connect chart sets this default to true](https://github.com/rstudio/helm/blob/228a98e729f9e2d0de9c16898f79dfe73759ad01/charts/rstudio-connect/values.yaml#L282).

### DIFF
--- a/charts/rstudio-workbench/README.md
+++ b/charts/rstudio-workbench/README.md
@@ -150,7 +150,7 @@ Alternatively, database passwords may be set during `helm install` with the foll
 - In most places, we opt to pass Helm values directly into ConfigMaps. We automatically translate these into the
   valid `.ini` or `.dcf` file formats required by Workbench.
   - Those configuration files and their mount locations are covered in the [Configuration files](#configuration-files) section below.
-- If you need to modify the jobs (sessions) launched by Workbench, use the `launcher.templates` as described in the [Launcher Templates](#launcher-templates) section below.
+- If you need to modify the jobs (sessions) launched by Workbench, use `launcher.templateValues` as described in the [Launcher Templates](#launcher-templates) section below.
 - The prestart scripts for Workbench and Posit Job Launcher are highly customized to get the service account information off of the Workbench pod for use in launching jobs.
 
 ## Configuration files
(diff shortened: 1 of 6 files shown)


charts/rstudio-workbench/README.md
```diff
@@ -398,37 +388,29 @@ some-key: value1,value2,value3,value4
...
-            name: imagePullPolicy
-          - target: "/spec/template/spec/imagePullSecrets"
-            json:
-              - name: my-pull-secret
-            name: imagePullSecrets
-        container-images:
-          - "one-image:tag"
-          - "two-image:tag
+launcher:
+  templateValues:
```
GCRev, line 408: Maybe this is me not understanding Kubernetes, but we call out `launcher.templates` in all the documentation, but all the examples use `templateValues` instead. Is that just how it is?

tnederlof: Good catch, it should be `launcher. templateValues` I just pushed the change

GCRev (approved):
LGTM and we hope people read the changelog

PR #662, opened 2025-05-02
ci: Add workflow to assign product-specific issues to their project boards

### LINKED ISSUE #658: ci: add workflow to assign issues to team's projects based on label
So that the product teams get visibility on relevant helm issues

nealrichardson:
Fixes #658

### DIFF
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -23,7 +23,6 @@ jobs:
       - name: Generate token for GitHub App in the posit-dev org
         id: generate-token-posit-dev
         uses: actions/create-github-app-token@v2
-        if:
         with:
           app-id: 1064714
           private-key: ${{ secrets.POSIT_PLATFORM_PROJECTS_PEM }}
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -1,0 +1,49 @@
+# When an issue is labeled, add to a project
+on:
+  issues:
+    types:
+    - labeled
+    - opened
+  # This is just for testing, comment out before merging
+  pull_request:
+    types:
+    - labeled
+
+jobs:
+  issues:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate token for GitHub App
+        id: generate-token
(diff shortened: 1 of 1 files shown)


.github/workflows/issues.yml
```diff
@@ -13,37 +13,46 @@ jobs:
...
+        id: generate-token-rstudio
         uses: actions/create-github-app-token@v2
         with:
           app-id: 1064714
           private-key: ${{ secrets.POSIT_PLATFORM_PROJECTS_PEM }}
+          owner: rstudio
+      - name: Generate token for GitHub App in the posit-dev org
+        id: generate-token-posit-dev
+        uses: actions/create-github-app-token@v2
+        if:
```
jspiewak, line 26: does empty `if:` work?

nealrichardson: lol, apparently so! I started to check to only do this if the label is for connect but the VS Code plugin for GHA didn't like the syntax, and I thought I deleted this. I'll clean it up.

nealrichardson:
@jspiewak I rewrote this a bit, PTAL. Here are the test cases: 

* Adding to a project in the rstudio org: https://github.com/rstudio/helm/actions/runs/14843562373/job/41671993954?pr=662
* Adding to a project in posit-dev: https://github.com/rstudio/helm/actions/runs/14843622340/job/41672179912?pr=662
* No-op when adding any other label: https://github.com/rstudio/helm/actions/runs/14843638294/job/41672229134?pr=662

There seems to be a quirk in `add-to-project` related to issues in projects across organizations, something GitHub generally struggles with. `add-to-project` is fine to re-run if the issue is already in the project generally, but across orgs, it appears to fail. But since this workflow now just triggers on the `labeled` event, and it doesn't pay attention to labels that are already set, you would only hit that failure mode if someone added the `team: connect` label (which adds to the project), removes the label, and adds the label back again. That seems unlikely, and anyway it's not a problem if it happens, it just means the workflow errors but nothing is destroyed as a result.

jforest (approved):
🤘

PR #671, opened 2025-05-15
Add Chronicle agent builtin sidecar definitions to Connect and Workbench

ianpittwood:
- Add configurable `initContainers` definition for Chronicle agent to Workbench and Connect charts.
- Bump `rstudio-library` to 0.1.32 in Workbench and Connect dependencies.
- Add information on Chronicle to READMEs.


charts/rstudio-workbench/values.yaml
```diff
@@ -63,6 +63,30 @@ strategy:
...
+    repository: rstudio/chronicle-agent
+    # -- A tag to use for the Chronicle Agent image. If not set, the chart will attempt to look up the version of the deployed Chronicle server in the current namespace.
+    tag: ""
+    # -- The pull policy for the Chronicle Agent image
+    imagePullPolicy: IfNotPresent
+  # -- The address for the Chronicle server. If not set, the chart will attempt to look up the address of the Chronicle Server in the release namespace or the serverNamespace if provided.
+  serverAddress: ""
+  # -- The namespace for the Chronicle server. If not set, the chart will attempt to look up the address of the Chronicle Server in the release namespace.
+  serverNamespace: ""
+
```
t-margheim, line 89: There should be a way to provide a Workbench API key here as well. This API key is optional for agents running alongside Workbench, but we need to provide a mechanism for a user who wants to to provide one.

charts/rstudio-connect/values.yaml
```diff
@@ -261,6 +261,33 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+# -- Settings for the Chronicle Agent sidecar container
+chronicleAgent:
```
markrtucker, line 265: I would have expected `environment` to be an option here. (This is related to the perhaps-poorly-named "environment" option for the agent: https://docs.posit.co/chronicle/appendix/library/advanced-agent.html#environment).

Maybe this is defined somewhere else?

charts/rstudio-workbench/values.yaml
ianpittwood: I added `chronicleAgent.workbenchApiKey` to provide an override.

charts/rstudio-connect/values.yaml
ianpittwood: I think since this comment I added `env` for this purpose.

charts/rstudio-workbench/values.yaml
```diff
@@ -63,6 +63,30 @@ strategy:
     maxSurge: "100%"
     maxUnavailable: 0
 
+# -- Settings for the Chronicle Agent sidecar container
+chronicleAgent:
```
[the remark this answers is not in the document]
ianpittwood: Same here, see `env`.

charts/rstudio-workbench/README.md
```diff
@@ -413,6 +413,54 @@ launcher:
...
+    env:
+      - name: CHRONICLE_SERVER_ADDRESS
+        value: "http://<address>"
+```
+
+For more information on Posit Chronicle, see the [Chronicle documentation](https://docs.posit.co/chronicle/).
+
+### Chronicle Workbench API Key
+
+In most cases where the Chronicle agent is deployed as a sidecar to Workbench, the agent can provision its own API key.
```
t-margheim, line 452: I don't think this is true, but maybe I've missed something. Do we automatically generate an API key as part of this chart now?

ianpittwood: Perhaps I misunderstand the process here? I thought Chronicle magically provisioned its own API key. Does it simply not require one for Workbench in most cases for this setup with a sidecar?

t-margheim: Yeah, sorry we should probably have been more explicit about this. Since [the Workbench API is still a "preview" feature](https://docs.posit.co/ide/server-pro/admin/workbench_api/workbench_api.html), we don't require it to be enabled in order to run Chronicle. Instead, if a user doesn't supply a Workbench API key, we just don't turn on Workbench API scraping in the agent.

charts/rstudio-connect/README.md.gotmpl
```diff
@@ -64,6 +64,66 @@ Alternatively, database passwords may be set during `helm install` with the foll
...
+      secretKeyRef:
+        name: <secret-name>
+        key: <key-name>
+```
+
+Due to the way Connect manages its API keys, it is currently not possible to provision an API key automatically for the
+Chronicle agent at the time of deployment. To workaround this issue in a fresh deployment, you can initially leave
+the API key unset for the Chronicle agent, deploy the chart, create an administrator API key, and then provision a
+secret with the API key. Once the secret is created, the value of `chronicleAgent.connectApiKey.secretKeyRef`
+can be set and the chart can be upgraded to include the new value.
```
bschwedler: @t-margheim @markrtucker 
Should we include similar language in the README for the Workbench chart?
Since the Workbench API is in preview, I didn't know if that changes the decision.

charts/rstudio-workbench/README.md
```diff
@@ -413,6 +413,62 @@ launcher:
...
+
+### Chronicle Workbench API Key
+
+> [!WARNING]
+> The Workbench API is currently in preview. See
+> [the Workbench documentation](https://docs.posit.co/ide/server-pro/admin/workbench_api/workbench_api.html) for more
+> information.
+
+The Chronicle agent can be configured to scrape the Workbench API for additional data. To do this, you must
+provide the Chronicle agent with a Workbench API key. This can be done by setting `chronicleAgent.workbenchApiKey`:
```
jforest, line 458: Is this the same as connect, must be created first and then a second deployment with the key defined as a secret?  Does that key need to be manually pushed to the k8s cluster as a secret before this will work?

charts/rstudio-connect/README.md.gotmpl
```diff
@@ -64,6 +64,66 @@ Alternatively, database passwords may be set during `helm install` with the foll
 
 `--set config.Postgres.Password="<YOUR_PASSWORD_HERE>"`
 
+## Chronicle Agent
+
+This chart supports use of a sidecar Chronicle agent to report data to a Chronicle server. The agent can be enabled
```
bschwedler, line 69: Should we link to the agent docs here?  We have a reference further down to the docs, but it may be good introduce earlier.
```suggestion
This chart supports use of a sidecar [Chronicle agent](https://docs.posit.co/chronicle/) to report data to a Chronicle server. The agent can be enabled
```

charts/rstudio-workbench/README.md.gotmpl
```diff
@@ -358,6 +358,62 @@ launcher:
           effect: "NoSchedule"
 ```
 
+## Chronicle Agent
+
+This chart supports use of a sidecar Chronicle agent to report data to a Chronicle server. The agent can be enabled
```
bschwedler, line 363: Same as the connect chart: should we link earlier?
```suggestion
This chart supports use of a sidecar [Chronicle agent](https://docs.posit.co/chronicle/) to report data to a Chronicle server. The agent can be enabled
```

bschwedler (approved):
This is looking great! I especially like the auto-discovery capability.

@markrtucker @t-margheim Could you please look over these changes once more?

charts/rstudio-connect/values.yaml
markrtucker: I don't think this is quite the same thing.

We have a gcfg variable named `Agent.Environment`. This is optional, but intended to allow our users to "tag" metrics with the deployment environment name (e.g., "test" vs "prod").

We expect most small customers won't care about this, but most large customers will heavily rely on this option.

I would expect that where we have `serverAddress` and `connectApiKey`, that we would have something like `chronicleEnvironment` (or perhaps some better name???). Conceptually, these variables are all the same kind of thing (configuration for the agent).

charts/rstudio-workbench/values.yaml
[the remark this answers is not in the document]
markrtucker: Same comment as above

charts/rstudio-connect/values.yaml
bschwedler: Ah! So this is a difference between [environment variable over-rides](https://docs.posit.co/chronicle/appendix/library/advanced-agent.html#environment-variables) and the `Agent.Environment` configuration value. Thanks for the clarification.

ianpittwood: I created `agentEnvironment` for this purpose.

charts/rstudio-workbench/values.yaml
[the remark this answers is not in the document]
ianpittwood: See above.

charts/rstudio-connect/README.md.gotmpl
bdeitte: I think that makes sense to include up here too

charts/rstudio-connect/README.md.gotmpl
```diff
@@ -64,6 +64,66 @@ Alternatively, database passwords may be set during `helm install` with the foll
...
+by setting `chronicleAgent.enabled=true`.
+
+By default, the chart will attempt to lookup an existing Chronicle server deployed in the release namespace. The
+searched namespace can be changed setting `chronicleAgent.serverNamespace`. If a server exists, it will set the
```
bdeitte, line 73:
```suggestion
searched namespace can be changed by setting `chronicleAgent.serverNamespace`. If a server exists, it will set the
```

charts/rstudio-workbench/README.md
```diff
@@ -413,6 +413,62 @@ launcher:
           effect: "NoSchedule"
 ```
 
+## Chronicle Agent
+
+This chart supports use of a sidecar Chronicle agent to report data to a Chronicle server. The agent can be enabled
+by setting `chronicleAgent.enabled=true`.
+
+By default, the chart will attempt to lookup an existing Chronicle server deployed in the release namespace. The
+searched namespace can be changed setting `chronicleAgent.serverNamespace`. If a server exists, it will set the
```
bdeitte, line 422:
```suggestion
searched namespace can be changed by setting `chronicleAgent.serverNamespace`. If a server exists, it will set the
```

charts/rstudio-workbench/README.md
ianpittwood: Workbench's API key is more optional since it already exposes much more metric data over Prometheus than Connect currently does. The agent will not crash loop if it's not provided the API key right off the bat so users are safe doing their initial deployment in this case without needing to be concerned about the API key being set.

charts/rstudio-connect/README.md
```diff
@@ -124,6 +124,66 @@ Alternatively, database passwords may be set during `helm install` with the foll
 
 `--set config.Postgres.Password="<YOUR_PASSWORD_HERE>"`
 
+## Chronicle Agent
+
+This chart supports use of a sidecar Chronicle agent to report data to a Chronicle server. The agent can be enabled
+by setting `chronicleAgent.enabled=true`.
+
+By default, the chart will attempt to lookup an existing Chronicle server deployed in the release namespace. The
+searched namespace can be changed setting `chronicleAgent.serverNamespace`. If a server exists, it will set the
```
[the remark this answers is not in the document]
ianpittwood: Good catch, fixed!

bschwedler (approved):

Issue #436, opened 2023-11-02
Change launcher for Connect chart to default true


tnederlof:
In the Connect documentation, we tell people to set `launcher.enabled` to `true`, however, the helm chart default is `false`. This is confusing to customers who find the helm chart and deploy Connect without seeing the Connect Admin Guide docs as it deploys not using OHE.

https://docs.posit.co/connect/admin/getting-started/off-host-install/configure-helm-chart/#step-1-create-your-initial-values.yaml-file

tnederlof:
Also should switch `securityContext.privileged: false` along with the launcher change

PR #672, opened 2025-05-16
PCT: Change the Connect default to use OHE

tnederlof:
This PR changes the Connect chart to run in OHE mode by default instead of the current default (service and content runs in the same pod(s)). I think this default makes more sense as its the way we at Posit advise customers to run Connect in Kubernetes and it is the only Kubernetes configuration [documented in the admin guide](https://docs.posit.co/connect/admin/getting-started/off-host-install/).

This is also important for customers who have [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/) or other mechanisms set to disallow privileged execution. The current chart runs with `securityContext.privileged: true`, which causes issues for these customers until they know to switch this setting. This PR will eliminate that need.

Bumped the minor version and added a section to the README template to make it more prominent the potentially breaking change (for customers who are not running in OHE).

Once this merges, the documentation here can be simplified (don't need to explicitly have folks add `launcher.enabled: true`. https://docs.posit.co/connect/admin/getting-started/off-host-install/configure-helm-chart/

Closes: https://github.com/rstudio/helm/issues/436


tnederlof:
It looks like the install in CI is failing because there is no PVC. I am not familiar with the testing infrastructure. Is there a way to mock up storage in these tests? It seems this error is reasonable if a customer is trying to setup without sharedStorage.

```
Error: template: rstudio-connect/templates/NOTES.txt:46:6: executing "rstudio-connect/templates/NOTES.txt" at <fail "\n\nWhen launcher is enabled, persistent storage must be provided.\nThis is usually done via a PersistentVolumeClaim (PVC) with `sharedStorage.create=true`, although there are other options.">: error calling fail: HELM_ERR_START

When launcher is enabled, persistent storage must be provided.
This is usually done via a PersistentVolumeClaim (PVC) with `sharedStorage.create=true`, although there are other options.HELM_ERR_END
```

bdeitte:
@jforest In the comment above, is this the part of the testing infra you worked on?

jforest:
> @jforest In the comment above, is this the part of the testing infra you worked on?

This is the install testing, yes.   This is triggered because the chart requires a pvc when `launcher.enabled` is `true`

https://github.com/rstudio/helm/blob/main/ci/rstudio-connect/install/license-file-values.yaml is the only set of install testing configs.  Each file in that directory will be used for a different install test.  These tests are run by the [chart testing tool](https://github.com/helm/chart-testing). The lint tests are also run by that tool.

Would you like to set up a time to talk through how all the testing is set up?

tnederlof:
@jforest gave me a walkthrough of the testing and was able to adjust the CI run so it creates a PVC so tests pass now. This should be ready for review.

charts/rstudio-connect/README.md
```diff
@@ -43,6 +43,13 @@ To explore other chart versions, look at:
 helm search repo rstudio/rstudio-connect -l
 ```
 
+## Upgrade guidance
+
+### 0.8.0
+
+- When upgrading to version 0.8.0 or later, Connect now runs in [Off-Host Execution mode](https://docs.posit.co/connect/admin/getting-started/off-host-install/) by default
+- If you desire to run Connect not in Off-Host Execution mode, then set `securityContext.privileged: true` and `launcher.enabled: false`
```
jforest, line 51: 🤘

charts/rstudio-connect/NEWS.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.8.0
+
+- BREAKING: Connect now runs in [Off-Host Execution (OHE) mode](https://docs.posit.co/connect/admin/getting-started/off-host-install/) by default.
```
jforest, line 5: 🤘

bschwedler (approved):

bschwedler:
@aronatkins @dbkegley @zackverham 
Do you forsee any issues merging this PR?

dbkegley (approved):

PR #676, opened 2025-05-23
Use helmignore for test directories

bschwedler:
- **Update .helmignore to skip test directories.**
- **Move test files into chart directories**
- **Use helmignore for chart-testing and unittest**


jforest (approved):
I love the cleanup!  Simpler and easier to understand, thanks for taking this on!

PR #689, opened 2025-06-13
Update Workbench settings to reflect recent defaults: remove JupyterHub sessions + align VSCode / Positron user settings

fh-mthomson:
Sharing back a few VSCode / Positron settings we made to override those set by default in Posit's helm charts, to more closely align with recent Workbench defaults.

CI will likely fail, but feel free to take / leave any as you see helpful!


GCRev (approved):
LGTM

PR #693, opened 2025-07-03
Update Chronicle default persistence accessMode to ReadWriteMany

samcofer:
Discussion on this change happened internally, but the take-aways are this:

1. This change does not enable multiple Chronicle server replicas and doesn't enable any type of HA
2. This change is purely to allow Connect pods to run on other hosts while reporting on Chronicle data and makes the `values.yaml` enable this configuration as a default.


t-margheim (approved):
Thanks @samcofer!

PR #696, opened 2025-07-10
PWB: Fix README and example for using `imagePullSecrets` for sessions

tnederlof:
I noticed the current Workbench example and README provide the incorrect syntax for providing container registry credentials for session images. This PR fixes both the example and the section of the Workbench chart README.

This now lines up with the Connect example which is correct: https://github.com/rstudio/helm/blob/b83d01cfe3af84cbea06a95de73c1a1ff536140a/examples/connect/container-images/rstudio-connect-custom-image-private.yaml#L37

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-workbench
 description: Official Helm chart for Posit Workbench
-version: 0.9.6
+version: 0.9.7
 apiVersion: v2
 appVersion: 2025.05.1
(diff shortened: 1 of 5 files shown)


GCRev (approved):

tnederlof:
Thanks for taking a look, @zachhannum and @GCRev.

CI is getting stuck on installing the chart `posit-chronicle`, which is unrelated. Perhaps @rstudio/infraops knows more?

I'll make a PR to the Workbench docs because there is an example in there as well (I added these originally at a similar time and messed up the syntax).

t-margheim:
@tnederlof: I'm taking a look at why the `posit-chronicle` chart is causing issues in CI. I'll let you know when we find a fix.

PR #698, opened 2025-07-22
Add ephemeral storage support to the Workbench Helm chart for session pods

samcofer:
This change allows administrators to define ephemeral-storage defaults for Workbench session containers!

### DIFF
--- a/charts/rstudio-workbench/files/job.tpl
+++ b/charts/rstudio-workbench/files/job.tpl
@@ -309,19 +309,21 @@ spec:
               {{- end }}
             {{- end }}
           {{- end }}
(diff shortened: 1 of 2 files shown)


zachhannum (approved):

samcofer:
@zachhannum Do you know why the checks are stuck? It won't let me merge at the minute, even though I see the checks completed here: https://github.com/rstudio/helm/actions/runs/16454882741

tnederlof:
The actions are weird. To get around it I usually pull down the changes made by the auto "Update helm-docs and README.md", change a character in the NEWS (or somewhere that doesn't trigger that action) then push back up and it will clear the checks. In this case I see there is no NEWS so not sure how to get around it...

PR #699, opened 2025-07-28
add resource limits to chronicle stateful set

markrtucker:


charts/posit-chronicle/NEWS.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.4.4
+- Add resource requests and limits to the chart.
```
ianpittwood, line 4: Should this be a patch? Are there any situations where this could potentially break something?

charts/posit-chronicle/values.yaml
```diff
@@ -78,6 +78,15 @@ pod:
     fsGroup: 1000
     fsGroupChangePolicy: "OnRootMismatch"
 
+resources:
+  requests:
+    memory: "2Gi"
+    cpu: "1000m"
+    ephemeralStorage: "100Mi"
```
ianpittwood, line 85: I don't have the whole context for these changes, but we should be sure that setting minimums and limits on CPU and ephemeral storage is actually necessary. I think including the option for users is nice though.

charts/posit-chronicle/values.yaml
```diff
@@ -78,6 +78,15 @@ pod:
...
+  limits:
+    memory: "3Gi"
+    cpu: "2000m"
+    ephemeralStorage: "200Mi"
```
ianpittwood, line 89: I imagine other charts are storing this at the root level, but I think it could be nested in `pod` for clarity if we also move the `resources` definition to the pod spec level in the `stateful-set.yaml` as mentioned above.

markrtucker: Moved to the `pod` level.

charts/rstudio-connect/values.yaml
```diff
@@ -286,6 +286,14 @@ chronicleAgent:
...
   # -- If true, the chart will attempt to lookup the Chronicle Server address and version in the cluster
   autoDiscovery: true
+  # -- Defines resources for the posit-chronicle-agent container
+  resources: {}
+    # requests:
+    #   memory: "256Mi"
+    #   cpu: "100m"
+    # limits:
+    #   memory: "512Mi"
+    #   cpu: "200m"
```
t-margheim, line 296: In the default case, are the agent and product sharing the same pool of memory? In other words, if the agent hit `OOM killed`, would that take down the colocated product as well?

markrtucker: I'm not completely sure about this. As I understand it, the pod is sharing resources, but with no limits defined, it will consume whatever it needs from the node (unless/until the node runs out of memory).

I know Ian talked about the initContainer having a separate lifecycle from the main container. But I don't know if that applies to OOM situations as well.

@ianpittwood - any thoughts here?

markrtucker:
> It does look like you'll need to push an empty commit in order to get CI to trigger.

Yeah - I'm not sure what was happening. CI was running (I was watching the actions directly). I'm not sure why the PR wasn't getting the status. It seems to be working now.

PR #701, opened 2025-07-31
Ephemeral storage fix

samcofer:
I received reports of errors occurring in internal environments because ephemeral storage was being set all the time, rather than only when defined in the `values.yaml` this PR appropriately wraps the ephemeralStorage definitions to set them only if they have a value.


zachhannum (approved):
Makes sense!

PR #702, opened 2025-08-06
Chronicle agent server address resolution fixes

ianpittwood:
- Skip calling `rstudio-library` when `chronicleAgent.autoDiscovery=false`. Require `chronicleAgent.serverAddress` be specified instead.
- Add validation step in `rstudio-library.chronicle-agent.serverAddress` to check if the lookup returns an empty value.
- Fail `rstudio-library.chronicle-agent.serverAddress` render if no server address can be resolved.

### DIFF
--- a/charts/rstudio-connect/Chart.yaml
+++ b/charts/rstudio-connect/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-connect
 description: Official Helm chart for Posit Connect
-version: 0.8.6
+version: 0.8.7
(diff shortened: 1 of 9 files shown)


matt-urbina:
@ianpittwood Updated this PR from main since it's sat awhile. Is there a way you would recommend I can test and validate these changes?

Here's the [slack thread](https://positpbc.slack.com/archives/C046857UA59/p1754492431742339) again incase you need to regain context!

PR #705, opened 2025-08-18
Only send Slack messages on main

bdeitte:
As discussed internally, we just want to see Slack messages for main failures now.  They will still of course still run, show up here, message people in their own way on failure, and stop people from merging when things fail.


bschwedler (commented):
Do we also want this change also to apply to [chart-doc.yaml](https://github.com/rstudio/helm/blob/main/.github/workflows/chart-doc.yaml)

PR #714, opened 2025-09-15
Add new workbench.

ssinnott:
Ok last part of the release (I think). Bumping the version in helm.

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,8 +1,8 @@
 name: rstudio-workbench
 description: Official Helm chart for Posit Workbench
-version: 0.9.11
+version: 0.9.12
 apiVersion: v2
-appVersion: 2025.05.1
+appVersion: 2025.09.0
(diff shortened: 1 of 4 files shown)


adamhigerd (approved):

charts/rstudio-workbench/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 name: rstudio-workbench
 description: Official Helm chart for Posit Workbench
-version: 0.9.11
+version: 0.9.12
```
GCRev, line 3: Should we be bumping this up by a more major amount, or is that not helm kosher?

GCRev (approved):
LGTM. Hopefully those settings are valid

zachhannum:
> LGTM. Hopefully those settings are valid

The new default settings come from the defaults defined in our docs: https://docs.posit.co/ide/server-pro/admin/positron_sessions/user_settings.html#default-settings.

charts/rstudio-workbench/Chart.yaml
zachhannum: Looks like the minor version is typically bumped if breaking changes to the chart itself are introduced, so I don't think that applies here.

PR #717, opened 2025-09-19
feat(workbench): turn on positron by default

fh-mthomson:
Now that Positron is GA as of 2025.09.0, would suggest updating `positron.conf` settings in the Helm chart accordingly!

Follow-on to :
* #631
* #689

Relates to #714 

Reference: https://docs.posit.co/ide/server-pro/admin/positron_sessions/user_settings.html

### DIFF
--- a/charts/rstudio-workbench/NEWS.md
+++ b/charts/rstudio-workbench/NEWS.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.9.12git 
+## 0.9.12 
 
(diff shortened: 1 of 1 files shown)


zachhannum:
Hey @fh-mthomson! Thanks for the contribution. I've gone ahead and folded in changes to enable positron by default as a part of the changes to bump the Workbench version in #714 so that all of those changes can land at once with a single chart version bump.

PR #723, opened 2025-10-09
Posit Workbench: Support defining sensitive config values via k8s secrets

kaypeter87:
This PR is copied over from https://github.com/rstudio/helm/pull/535 due to the inactivity of the original author. I have also added the suggested changes from the reviewer.

Applies to

- launcherPem
- secureCookieKey
- userPassword
- database.conf

When doing so, the `rstudio-secret` is skipped and individual mounts containing the above-mentioned secret values are injected.

This PR is breaking as the items listed are now maps and existing deployments would need to migrate from `launcherPem: <value>` to `launcherPem.value: <value>`.

@pat-s tested this in a deployment of theirs and it works as expected so far.

```yaml
secureCookieKey:
  existingSecret: secure-cookie-key
launcherPem:
  existingSecret: launcher-pem
config:
  database:
    conf:
      existingSecret: database-conf
userPassword:
  existingSecret: workbench-user-password
```

The advantage of this approach is that it gives power to the users and makes use of "simple" k8s secrets instead of relying on some chart magic which puts together a bundled secret from plain config values.

related https://github.com/rstudio/helm/issues/520 https://github.com/rstudio/helm/pull/493

### DIFF
--- a/charts/rstudio-connect/Chart.yaml
+++ b/charts/rstudio-connect/Chart.yaml
@@ -1,8 +1,8 @@
 name: rstudio-connect
 description: Official Helm chart for Posit Connect
-version: 0.8.9
+version: 0.8.11
 apiVersion: v2
-appVersion: 2025.09.0
+appVersion: 2025.09.1
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
 home: https://www.rstudio.com
 sources:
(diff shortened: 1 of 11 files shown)


kaypeter87:
Closing as the original author of https://github.com/rstudio/helm/pull/535 is currently engaged and active on the original PR.

PR #730, opened 2025-10-31
Bump workbench version to 2025.09.2

jeffvroom:


skyeturriff:
re-running checks as those docker images should be available now

PR #732, opened 2025-11-07
Use 4x runners since we keep running out of space

jforest:


jforest:
> 🤞🏼

The install tests worked on the PR so merging!

Issue #733, opened 2025-11-07
Add explicit support for client secrets by volume mount.


ssinnott:
## Description

Similar to this PR: https://github.com/rstudio/helm/pull/726

We suggest,
1) Storing secrets as kubernetes secrets.
2) Mounting the secrets into the container by volume mount.

We have a few secrets which are not explicitly supported in helm in this way - including the openid client secret. So we should add support for that as well.

fh-albert-lee:
I drafted https://github.com/rstudio/helm/pull/736 to address this issue. Its not specific to the openid client secret, but allows adding it via an `existingSecrets` list that will get projected to the secrets volume in the same pattern as #726.

PR #738, opened 2025-11-18
Fh al wb conditional secret fields

GCRev:
From: https://github.com/rstudio/helm/pull/736

Related to #733 and #726

For the `rstudio-workbench` Helm chart, add support for referencing externally managed Kubernetes Secrets via an `existingSecrets` Helm value.

- The chart now conditionally excludes `launcher.pem` and `secure-cookie-key` from the chart-managed secret (`configmap-secret.yaml`) when `existingSecret` is configured for each value. This avoids duplication, ensures the existing secret is the single source of truth, and prevents exposure in version control if using this chart via the [rendered manifests](https://akuity.io/blog/the-rendered-manifests-pattern) pattern.
- Added `config.existingSecrets` to support projecting arbitrary external Kubernetes Secrets into `/mnt/secret-configmap/rstudio/` (e.g., for `openid-client-secret` and other sensitive configurations).
- Added documentation in README covering both features with examples.
- Added 4 new test cases covering the new functionality.

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-workbench
 description: Official Helm chart for Posit Workbench
-version: 0.10.1
+version: 0.10.2
 apiVersion: v2
 appVersion: 2025.09.2
 icon: https://rstudio.com/wp-content/uploads/2018/10/RStudio-Logo-Flat.png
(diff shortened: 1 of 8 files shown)


GCRev:
@jforest it was merged to this non-forked branch so that GHAs could run on it without blowing up. The code-owner policy still requires a review before merging, but because I created this branch, I cannot also approve it. We need some more code owners :(

Issue #753, opened 2026-01-20
Allow more complex maps in `securityContext` than just key-value pairs


jacpete:
## Description
This was discovered through Posit support ticket 125929. They brought up two potential issues in the Connect Helm chart in the job.tpl file.  Both issues are found in this section of code: https://github.com/rstudio/helm/blob/c6c1b08b849af5c4fbfbf90119fee6eb27cb69fc/charts/rstudio-connect/files/job.tpl#L133-L146

1.  The current method for templating the `securityContext` section does not allow for more complex mapping structures in the values.yaml than key-value pairs currently. Our code is currently: https://github.com/rstudio/helm/blob/c6c1b08b849af5c4fbfbf90119fee6eb27cb69fc/charts/rstudio-connect/files/job.tpl#L141-L146
but this fails in a situation where the you have more than key-value pairs in your values.yaml:
  ```yaml
  securityContext:
    appArmorProfile:
      type: RuntimeDefault
    seccompProfile:
      type: RuntimeDefault
    capabilities:
      drop:
        - ALL
    allowPrivilegeEscalation: false
    privileged: false
  ```
Example error `"json: cannot unmarshal string into Go struct field PodSecurityContext.spec.template.spec.securityContext.appArmorProfile of type v1.AppArmorProfile"`

2. The second issue is that the mergeOverwrite only happens if `.Job.container.supplementalGroupIds` exists.
https://github.com/rstudio/helm/blob/c6c1b08b849af5c4fbfbf90119fee6eb27cb69fc/charts/rstudio-connect/files/job.tpl#L133-L140

## Potential fix
I believe the following replacement for that section of code could fix both these issues. First, by changing the `range` function to uses `toYaml` instead when expanding the `$securityContext` mapping. And second, moving the  `mergeOverwrite` line out of the `.Job.container.supplementalGroupIds` if statement and into the `$securityContext` if statement instead.
Original: https://github.com/rstudio/helm/blob/c6c1b08b849af5c4fbfbf90119fee6eb27cb69fc/charts/rstudio-connect/files/job.tpl#L133-L146
Replacement: 
```tpl
      {{- if .Job.container.supplementalGroupIds }}
        {{- $groupIds := list }}
        {{- range .Job.container.supplementalGroupIds }}
          {{- $groupIds = append $groupIds . }}
        {{- end }}
        {{- $_ := set $securityContext "supplementalGroups" (cat "[" ($groupIds | join ", ") "]") }}
      {{- end }}
[...]

PR #755, opened 2026-01-20
potential fix for #753

jacpete:
This a pull request for the fix suggested in and for issue #753. 

I believe the following replacement for that section of code could fix both these issues. First, by changing the `range` function to uses `toYaml` instead when expanding the `$securityContext` mapping. And second, moving the  `mergeOverwrite` line out of the `.Job.container.supplementalGroupIds` if statement and into the `$securityContext` if statement instead.

Noting that this may also be something that needs applied to other job.tpl files as well if approved. As an example:
https://github.com/rstudio/helm/blob/c6c1b08b849af5c4fbfbf90119fee6eb27cb69fc/charts/rstudio-workbench/files/job.tpl#L133-L146


lucasrod16:
Would you mind sharing some artifacts that verify the changes have been tested and work as expected? These changes are a bit trickier to test since this data is templated by the launcher at runtime and can't be easily validated with `helm lint`, `helm template` or unit tests.

Specifically, it would be helpful to see a launcher job successfully running with nested `securityContext` objects (e.g., `appArmorProfile`, `seccompProfile`).

jacpete:
@lucasrod16 Thanks! Looks like all the checks passed now.

PR #767, opened 2026-01-28
Fix Chart icons

bdeitte:
There is an artifact hub message that comes to me every helm deploy that I filter out and which I happened to see recently. I asked Claude to fix it for me, which led to seeing that these icons were no longer available. I also added a starter CLAUDE.md file

### DIFF
--- a/charts/rstudio-connect/README.md
+++ b/charts/rstudio-connect/README.md
@@ -30,11 +30,11 @@ To ensure reproducibility in your environment and insulate yourself from future 
 
 ## Installing the chart
 
-To install the chart with the release name `my-release` at version 0.8.21:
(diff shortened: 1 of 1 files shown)


GCRev (approved):
These changes look reasonable.

jforest:
I know the changes are super minimal, but I think we should update the NEWS.md for each of the charts to explain why the versions have been bumped.

examples/rbac/rstudio-launcher-rbac-0.2.25.yaml
```diff
@@ -0,0 +1,88 @@
+---
```
jforest: Is this an intentional addition?

Issue #505, opened 2024-05-14
Add /docs readme for updating and contributing to the Helm docs


AshleyHenry15:
It isn't clear what files actually "do the editing" for the Helm docs.

Probably should add to the readme that to edit the docs for Workbench, Connect, PPM, and Chronicle Overview pages, it requires a combo update of the following:

- /helm/charts/_templates.gotmpl
- charts/posit-chronicle/README.md.gotmpl
- charts/rstudio-connect/README.md.gotmpl
- charts/rstudio-pm/README.md.gotmpl
- charts/rstudio-workbench/README.md.gotmpl

And to preview changes:
```
quarto preview --render html
```

In addition, we should also add how to edit the Examples doc for each product as well.

Issue #62, opened 2021-10-19
Figure out how to do `helm-docs` action in PRs from public


colearendt:
Unfortunately `helm-docs` checks out a detached HEAD, so `helm-docs` cannot push changes, and the RBAC updates cannot pull changes with `git pull`. More advanced `git` CLI behavior may be necessary.

https://github.com/rstudio/helm/runs/3944250635

https://github.com/rstudio/helm/runs/3944247685

colearendt:
We have improved this with the `pr-fetch` and `pr-push` actions in `r-lib` (https://github.com/r-lib/actions). However, this is not currently possible because PRs from outside forks get a "read only" token for the repo.

To simplify this, we have added a `just` target that allows generating docs: `just docs`. Improving dev documentation (#72) will help us make this clear to contributors

PR #772, opened 2026-01-30
Update CI to detect uncommitted changes

dbkegley:
Our CI workflows currently attempt to generate docs and commit the result directly to the PR branch. This workflow prevents subsequent CI checks from running because GHA cannot trigger itself in PRs.

This change updates our workflows to force developers to generate docs locally and commit the result, avoiding the annoying necessity to create an `--alow-empty "trigger ci"` commit.

CI will now fail if uncommitted docs changes are detected. These changes are usually either an updated `Chart.lock` or `README.md`. They can be updated with `just update-lock` and `just docs` respectively.

Example docs failure:

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,15 +22,17 @@ We'll try to be as responsive as possible in reviewing and accepting pull reques
 
 ## Assumptions / Common Dev Workflows
 
+- Use `just docs` to generate the READMEs. READMEs are generated in CI by
+  [Go templating](./charts/_templates.gotmpl) and `helm-docs`.
+  - Run `just setup` to install `helm-doc`.
+- CI requires that the chart version get bumped for any change in the directory.
+- Use `just update-lock` to update all chart's lockfiles.
 - Changes to the `rstudio-library` chart will update all downstream charts at
   the same time (via the `file://` syntax in `Chart.yaml`)
 - CI only runs on local branches (i.e. not from forks). This can make
   evaluating code from contributors tricky. By creating a duplicate branch
   locally, we can "trick" CI into running on the same commits
-- CI requires that the chart version get bumped for any change in the directory
   (including README)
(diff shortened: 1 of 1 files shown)


dbkegley:
> > This change updates our workflows to force developers to generate docs locally and commit the result, avoiding the annoying necessity to create an `--alow-empty "trigger ci"` commit.
> 
> If the option is generate something locally and commit, versus trigger CI with an empty commit I think I would prefer the later during the release process. It is quite nice to not rely on anything being setup locally.

IMO this makes the product releases easier to automate because it won't require developers to go push an empty commit to unblock CI during the release. There would be no need to run anything locally because we can automate `just docs` during our release process.

For outside contributors we currently don't allow CI to run at all (https://github.com/rstudio/helm/issues/62) which is preventing community engagement. It has also proven challenging in the past for contributors to know how to make simple changes because CI gets stuck and it's not obvious how to unstick it.

https://github.com/rstudio/helm/issues/505

tools/dev/check-uncommitted-changes.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+if git status --porcelain=1 2>&1 | grep -E "^.|\bwarning\b" | grep -vE "\.gitignore"; then
+	echo "Error: uncommitted changes detected."
+	exit 1
+fi
```
lucasrod16: It would be helpful to log what the uncommitted changes were

lucasrod16:
> IMO this makes the product releases easier to automate because it won't require developers to go push an empty commit to unblock CI during the release. There would be no need to run anything locally because we can automate just docs during our release process.

Oh I see. I didn't realize that was the intended workflow. That makes sense to me

bschwedler (approved):
I like this approach.  This moves the responsibility of updating the docs to the PR creation process.

PR #786, opened 2026-02-24
Add direct Kubernetes off-host execution support to Connect chart

dbkegley:
This PR adds support for an alternative Kubernetes backend for off-host execution (OHE) in the Connect chart. Requires Connect version 2026.04.0 or later.

- Add `backends.kubernetes.enabled` and related values for configuring content Jobs and Services using standard Kubernetes manifests (`defaultResourceJobBase` / `defaultResourceServiceBase`)
- Add mutual-exclusion validation: `launcher.enabled` and `backends.kubernetes.enabled` cannot both be true
- Add dedicated least-privilege RBAC for the direct Kubernetes runner, with automatic ClusterRole creation for NodePort service types
- Auto-generate init container and runtime volume wiring in the job base, with fail-fast validation for reserved mountPath conflicts
- Add upgrade guide and before/after example values in `examples/connect/upgrade-launcher-to-kubernetes/`
- Bump rstudio-connect chart version to 0.9.0
- Add helm-unittest coverage for RBAC, configmap-prestart gating, init image tag precedence, and mount path validation

## Review guidance

Start with `examples/connect/upgrade-launcher-to-kubernetes/` — the upgrade guide (`launcher-to-kubernetes.qmd`) and the before/after values files are what customers will use to adopt this feature. Feedback on clarity, completeness, and whether the field mapping makes sense is the most valuable review input.

For the chart changes, the key areas are:
- `values.yaml` — the `backends.kubernetes` values surface and its documentation comments
- `_helpers.tpl` — the `rstudio-connect.jobBase` template that auto-generates the job manifest with init container, runtime volume, and mountPath conflict validation
- `rbac.yaml` — the direct Kubernetes runner RBAC is intentionally more restrictive than the Launcher RBAC (e.g., no `services/delete` — cleanup happens via Job cascade)

The test file (`tests/kubernetes_test.yaml`) is large but self-contained — skim the test names to see what's covered.


charts/rstudio-connect/values.yaml
```diff
@@ -312,6 +312,58 @@ chronicleAgent:
...
+    # -- The repository to use for the Content InitContainer image
+    repository: ghcr.io/rstudio/rstudio-connect-content-init
+    # -- A tag prefix for the Content InitContainer image (common selections: jammy-, ubuntu2204-). Only used if tag is not defined
+    tagPrefix: ubuntu2204-
+    # -- Overrides the image tag whose default is the chart appVersion.
+    tag: ""
+  # -- defaultJobOverlay contains the Kubernetes Job definition which is used as an overlay "base" when launching a content job
+  # in Kubernetes. Conceptually this is a similar to a Kustomize base. Connect then applies any required job configuration
+  # on-top of the overlay base to produce a final job definition before submitting the Job to Kubernetes.
+  # https://docs.posit.co/connect/__unreleased__/admin/appendix/off-host/direct-runner/kubernetes-job-overlays.html
```
jspiewak: This link will need to be updated prior to merging.

charts/rstudio-connect/values.yaml
```diff
@@ -312,6 +312,58 @@ chronicleAgent:
   #    ([reference](https://docs.posit.co/chronicle/appendix/library/advanced-agent.html#environment))
   agentEnvironment: ""
 
+kubernetes:
```
jspiewak: Is everyone happy with a value namespace of "kubernetes"?

charts/rstudio-connect/README.md
```diff
@@ -248,6 +248,14 @@ The Helm `config` values are converted into the `rstudio-connect.gcfg` service c
...
...Overlay | object | `{"apiVersion":"batch/v1","kind":"Job","spec":{"template":{"spec":{"containers":[{"name":"connect-content","volumeMounts":[{"mountPath":"/opt/rstudio-connect/R","name":"rsc-volume","subPath":"R"},{"mountPath":"/opt/rstudio-connect/python","name":"rsc-volume","subPath":"python"},{"mountPath":"/opt/rstudio-connect/scripts","name":"rsc-volume","subPath":"scripts"},{"mountPath":"/opt/rstudio-connect/ext","name":"rsc-volume","subPath":"ext"}]}],"initContainers":[{"name":"connect-content-init","volumeMounts":[{"mountPath":"/mnt/rstudio-connect-runtime/","name":"rsc-volume"}]}],"volumes":[{"emptyDir":{},"name":"rsc-volume"}]}}}}` | defaultJobOverlay contains the Kubernetes Job definition which is used as an overlay "base" when launching a content job in Kubernetes. Conceptually this is a similar to a Kustomize base. Connect then applies any required job configuration on-top of the overlay base to produce a final job definition before submitting the Job to Kubernetes. https://docs.posit.co/connect/__unreleased__/admin/appendix/off-host/direct-runner/kubernetes-job-overlays.html https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#bases-and-overlays |
```
samcofer: Why is there a populated defaultJobOverlay and not one for defaultServiceOverlay? 

Are the values you have listed there meaningful? ie, if I override this with a less complete object, does the content fail?

charts/rstudio-connect/values.yaml
```diff
@@ -312,6 +312,58 @@ chronicleAgent:
...
+      template:
+        spec:
+          containers:
+            - name: "connect-content" # well-known name
+              volumeMounts:
+                - mountPath: /opt/rstudio-connect
+                  name: rsc-volume
+          initContainers:
+            - name: connect-content-init # well-known name
+              # Note: Image is set dyncamically by kubernetes.defaultInitContainer
```
samcofer: Spell chekcer here

samcofer (commented):
One thing I'd love to see come along with these changes are some examples of side by side conversions from "old-launcher" to "new-launcher" for some key values. If we can get those added to examples/connect/launcher-migration, it would be a huge help for SE's when we're talking to customers about moving over.

charts/rstudio-connect/README.md
dbkegley: Yes, the defaults defined in the chart are required to configure the init-container. Any customization should be additive to the default values for the Job overlay.

The service overlay doesn't include any defaults because Connect will create a ClusterIP service by default. 

Note: We'll be renaming `defaultJobOverlay/defaultServiceOverlay` to be `defaultResourceJobBase/defaultResourceServiceBase` for clarity. Calling them overlays is a little misleading.

charts/rstudio-connect/values.yaml
```diff
@@ -312,6 +312,58 @@ chronicleAgent:
...
+    tagPrefix: ubuntu2204-
+    # -- Overrides the image tag whose default is the chart appVersion.
+    tag: ""
+  # -- defaultJobOverlay contains the Kubernetes Job definition which is used as an overlay "base" when launching a content job
+  # in Kubernetes. Conceptually this is a similar to a Kustomize base. Connect then applies any required job configuration
+  # on-top of the overlay base to produce a final job definition before submitting the Job to Kubernetes.
+  # https://docs.posit.co/connect/__unreleased__/admin/appendix/off-host/direct-runner/kubernetes-job-overlays.html
+  # https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#bases-and-overlays
+  defaultJobOverlay:
+    apiVersion: batch/v1
```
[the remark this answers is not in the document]
dbkegley: We'll provide a detailed upgrade guide that fully documents each field that users might have configured in the existing `launcher.templateValues` structure, along with where the values belong in the new job overlay.

We'll also provide a `SKILL.md` that customers can use to help automate the migration with claude or some other LLM.

> Given that the default pathing is deeper and requires a more detailed knowledge of Kubernetes nesting and yaml, it would be nice to provide that structure as a bridge

I'm not quite sure what you mean by this, could you explain? Are you saying that we should include a fully populated and commented job/pod manifest so the users can see the structure?

charts/rstudio-connect/values.yaml
samcofer: I'd ask you to think about it from an admin perspective. They have no clue that this doesn't _use_ something called launcher right? They don't/won't understand the difference between our internal "launcher" and this native approach. In their mind "launching" is a function, not a separate part of the code base. So I don't think this will be misleading to that audience.

That said, I'm totally cool with `offHost` I'd ask that we remove/eliminate the `kubernetes` concept/yaml level entirely. To my knowledge, Helm isn't applicable outside of Kubernetes, so why over specify it? Are you envisioning a future where we need an `offHost.slurm`? Since this a `helm` change, I'd treat it with that audience in mind. If there's a deeper meaning in the code, or elements of the future Connect roadmap that I'm not aware of, then maybe this distinction makes sense.

charts/rstudio-connect/tests/kubernetes-values.yaml
```diff
@@ -0,0 +1,24 @@
...
+                - mountPath: /opt/rstudio-connect
+                  name: rsc-volume
+          initContainers:
+            - name: connect-content-init
+              volumeMounts:
+                - mountPath: /mnt/rstudio-connect-runtime/
+                  name: rsc-volume
+          volumes:
+            - emptyDir: {}
+              name: rsc-volume
```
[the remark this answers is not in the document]
samcofer: Yup, that's pretty much what I'm talking about. I still think it would be a benefit to have an example in the examples folder that has both a full pod: and new launcher defined, with some comments explaining how they produce the same yaml.

charts/rstudio-connect/README.md
samcofer: I think that this behavior: 
> Connect will create a ClusterIP service by default.

And this behavior:

> the defaults defined in the chart are required to configure the init-container.

Seems counter intuitive right? Why does Connect do something without anything configured in the first case, but then require configuration to do the basics in the second case? My vote is to add the redundant serviceOverlay to the chart with the values that Connect is already defaulting too.

charts/rstudio-connect/NEWS.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.8.27
+
+- Adds support for a native-Go implementation of Off-Host Execution (OHE) mode. Set `kubernetes.enabled = true` to switch to the new implementation. See the [migration guide](<todo: insert migration guide docs link>) in the Connect admin documentation for guidance on migrating existing configurations.
```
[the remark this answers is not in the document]
SamEdwardes: I think we should add more details to the change log. Maybe showing a brief example of how you would do templating with Launcher and how you now do it with the new native Kubernetes.

charts/rstudio-connect/values.yaml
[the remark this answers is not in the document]
SamEdwardes: My preference would be to have little or no commented out YAML in the values.yaml. I always find that a bit confusing whether it was meant to be commented out or not. Instead, I'd rather have robust documentation. or comments that explain what can go in.

I also agree with Sam though that some of this documentation should be in the Helm Chart itself. Maybe some examples in the README.

SamEdwardes (commented):
Thank you for the work on this, David. I like this new approach and the flexibility that allows admins to make changes without us having to template in every single possible specification.

A left a few comments. Happy to discuss anything further.

charts/rstudio-connect/README.md
```diff
@@ -262,6 +262,14 @@ The Helm `config` values are converted into the `rstudio-connect.gcfg` service c
...
+| backends.kubernetes.defaultResourceJobBase | object | `{}` | defaultResourceJobBase is an optional Kubernetes Job definition used as the base when launching content jobs. The chart automatically adds the init container and runtime volume when backends.kubernetes.defaultInitContainer.enabled is true. Only set this if you need to customize the job (e.g., add sidecars, node selectors, tolerations). https://kubernetes.io/docs/concepts/workloads/controllers/job/ |
+| backends.kubernetes.defaultResourceServiceBase | object | `{}` | defaultResourceServiceBase contains the Kubernetes Service definition which is used as an overlay "base" when creating a content job's Service in Kubernetes. Conceptually this is a similar to a Kustomize base. Connect then applies any required Service configuration on-top of the overlay base to produce a final Service definition. https://docs.posit.co/connect/__unreleased__/admin/appendix/off-host/direct-runner/kubernetes-job-overlays.html https://kubernetes.io/docs/concepts/services-networking/service/ https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#bases-and-overlays |
```
dbkegley: TODO: link to updated Connect resource base docs
TODO: link to kubernetes service spec docs

dbkegley (commented):
Looks good so far.  A few things:

1. Let's come up with a new name besides "native" for this to avoid confusing it with native/local execution. 
2. We'll need to go back and fix some links in the values.yaml descriptions once we have docs to link to. To keep things simple maybe we should omit the Connect docs links for now and just link to the kubernetes docs.  We can come back later and add the Connect docs links to the top-level readme in the off-host section.
3. Let's add a callout in NOTES.txt when `launcher.enabled = true && appVersion >= 2026.04.0` alerting users that they should upgrade to the new implementation of off-host

charts/rstudio-connect/templates/_helpers.tpl
```diff
@@ -155,3 +178,106 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{ $key }}: {{ $value | quote }}
 {{ end }}
 {{- end -}}
+
+{{/*
+  Build the job resource base for the native Kubernetes runner.
+  Starts with the user-provided defaultResourceJobBase (or empty dict),
+  then adds the init container, content container, and shared volume
+  when defaultInitContainer.enabled is true.
+*/}}
```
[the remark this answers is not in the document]
lucasrod16: Absolutely

charts/rstudio-connect/templates/_helpers.tpl
```diff
@@ -90,6 +104,15 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
   {{- end }}
+  {{- /* remove empty DefaultResource*Base keys so they don't overwrite chart defaults */}}
+  {{- if hasKey $configCopy "Kubernetes" }}
+    {{- if not (dig "Kubernetes" "DefaultResourceJobBase" "" $configCopy) }}
+      {{- $_ := unset (index $configCopy "Kubernetes") "DefaultResourceJobBase" }}
+    {{- end }}
+    {{- if not (dig "Kubernetes" "DefaultResourceServiceBase" "" $configCopy) }}
+      {{- $_ := unset (index $configCopy "Kubernetes") "DefaultResourceServiceBase" }}
+    {{- end }}
+  {{- end }}
```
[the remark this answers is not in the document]
dbkegley: Ah, ty that makes sense!

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
```diff
@@ -0,0 +1,332 @@
+---
+category: "Upgrade"
+---
+
+# Transitioning to the direct Kubernetes runner
```
nihara-thomas: This page isn't very easy to find and doesn't really fit under the "examples" section. Could we move it to the left sidebar perhaps?

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
```diff
@@ -0,0 +1,332 @@
...
+
+This guide covers upgrading Posit Connect helm values from the Launcher (`launcher.enabled: true`) to the direct Kubernetes runner (`backends.kubernetes.enabled: true`).
```
nihara-thomas, line 7: Are we expecting customers to land here from the Connect docs where they might have already read a more detailed description of why they would want to move to the direct K8s runner?

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
```diff
@@ -0,0 +1,332 @@
...
+
+The direct Kubernetes runner manages content Jobs and Services using standard Kubernetes manifests, replacing the Launcher's template system.
+
+## Important notes
+
+:::{.callout-important}
+If you set `serviceAccountName` in the job base, you must label the service account for Connect to use it. Without this label, Connect falls back to the namespace default:
```
nihara-thomas: We should call out that customers need to apply the `connect.posit.co/service-account=true` label to any other service accounts that they want the apps to use as well.

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
```diff
@@ -0,0 +1,332 @@
...
+
+```bash
+kubectl label sa <service-account-name> connect.posit.co/service-account=true -n <namespace>
+```
+:::
+
+**Init container** -- The chart auto-generates a `connect-content-init` init container from `backends.kubernetes.defaultInitContainer`. You do not need to include it in `defaultResourceJobBase`. If you do include it, the chart updates its image, applies `defaultInitContainer` settings, and ensures the required volume mount. Custom init containers run alongside the auto-generated one.
```
nihara-thomas: Not really clear on if there's an action to be taken here. Could we rephrase this?

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
```diff
@@ -0,0 +1,332 @@
...
+
+## Values with no direct equivalent
+
+| Launcher value | Upgrade path |
+| --- | --- |
+| `launcher.customRuntimeYaml` | Use `executionEnvironments` instead. See [Execution environments](/charts/rstudio-connect/README.md#execution-environments). |
+| `launcher.additionalRuntimeImages` | Use `executionEnvironments` instead. |
+| `launcher.launcherKubernetesProfilesConf` | No equivalent. The direct Kubernetes runner does not use profiles. Similar customization can be achieved using `defaultResourceJobBase`. |
+
+## Minimal upgrade
```
nihara-thomas: I'm back to thinking this one should be up higher. Customers that don't modify `launcher.templateValues`, shouldn't have to read through the two sections above. They probably should know about service accounts even if they haven't set a default for all content in the helm charts.

Image

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
```diff
@@ -0,0 +1,332 @@
...
+```bash
+kubectl label sa <service-account-name> connect.posit.co/service-account=true -n <namespace>
+```
+:::
+
+**Init container** -- The chart auto-generates a `connect-content-init` init container from `backends.kubernetes.defaultInitContainer`. You do not need to include it in `defaultResourceJobBase`. If you do include it, the chart updates its image, applies `defaultInitContainer` settings, and ensures the required volume mount. Custom init containers run alongside the auto-generated one.
+
+**Shared storage** -- `sharedStorage.*` values (`create`, `mount`, `mountContent`, `name`, `subPath`, `storageClassName`) work identically for both modes. No changes needed.
+
+## Values with no direct equivalent
```
nihara-thomas: I think we should rephrase this. To me `no direct equivalent` reads like unsupported, which is not the case. How about something like this:
```suggestion
## Values with alternatives

These values do not have a direct equivalent but have alternative approaches in the Kubernetes runner.
```

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
```diff
@@ -0,0 +1,332 @@
...
+
+| Launcher value | Upgrade path |
+| --- | --- |
+| `launcher.customRuntimeYaml` | Use `executionEnvironments` instead. See [Execution environments](/charts/rstudio-connect/README.md#execution-environments). |
+| `launcher.additionalRuntimeImages` | Use `executionEnvironments` instead. |
+| `launcher.launcherKubernetesProfilesConf` | No equivalent. The direct Kubernetes runner does not use profiles. Similar customization can be achieved using `defaultResourceJobBase`. |
+
+## Minimal upgrade
+
+If you haven't customized `launcher.templateValues`, the minimal upgrade is a single change: set `launcher.enabled: false` and `backends.kubernetes.enabled: true`.
```
nihara-thomas: There's more in the table below? Should those be called out as "Optional" similar to what we have in the minimal example?

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
```diff
@@ -0,0 +1,332 @@
...
+          app.example.com/service: content
+        annotations:
+          app.example.com/tier: compute
+      spec:
+        type: ClusterIP
+```
+
+### Complete before/after reference
+
+For a complete example with all fields, see the full before/after files:
```
nihara-thomas, line 330: Should we also call out that anything that is supported by the K8s spec but not explicitly mapped out here is also supported?

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
```diff
@@ -0,0 +1,332 @@
+---
+category: "Upgrade"
```
nihara-thomas, line 2: Do we plan to update the `Configuring Posit Connect with Recommended Settings` doc to call out what a first time installation/configuration would look like in a separate PR?

The other docs under `Examples` also currently only reference launcher based values and can be confusing to someone not upgrading.

.gitignore
```diff
@@ -18,3 +18,6 @@ _publish.yml
 
 # helm packages
 *.tgz
+
+# local working notes
+notes.md
```
nihara-thomas: nit: should this be a local .gitignore thing instead of at the repo level?

nihara-thomas:
@samcofer we'd love your thoughts as well on the docs for anything that might be helpful to customers switching to direct k8s.

dbkegley (commented):
This looks good to me pending @nihara-thomas's feedback being addressed.

I can't approve because I created the PR initially so someone else will need to click the button

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
lucasrod16: Yes that's the plan. The Connect docs can cover the why and these helm chart docs cover the how. They will eventually both reference each other also, so hopefully that will help users have the full context.

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
lucasrod16: Added a note to the `Upgrading templateValues customizations` section

edavidaja (approved):

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
lucasrod16: Created #837 to capture this as a follow up

nihara-thomas (approved):
🚢 🚀 :shipit:

lucasrod16:
@bschwedler @ianpittwood @jspiewak can we get an approval from one of you so we can get this merged 🙏? We'd like to get this in ahead of our April release so we can do some further internal testing.

charts/rstudio-connect/values.yaml
```diff
@@ -471,8 +514,9 @@ config:
...
   Logging:
     ServiceLog: STDOUT
-    ServiceLogLevel: INFO    # INFO, WARNING or ERROR
-    ServiceLogFormat: TEXT   # TEXT or JSON
+    ServiceLogLevel: INFO  # INFO, WARNING or ERROR
+    ServiceLogFormat: TEXT  # TEXT or JSON
     AccessLog: STDOUT
     AccessLogFormat: COMMON  # COMMON, COMBINED, or JSON
   Launcher: {}
+  Kubernetes: {}
```
ianpittwood, line 522: It seems like a user could accidentally set `config.Kubernetes.Enabled` and `backend.kubernetes.enabled` to opposite values. The same is true for `config.Kubernetes.Namespace` and `backend.kubernetes.namespace`. 

Perhaps we should add a check to fail the deployment if the two are in conflict or enforce that only one section's value can be set.

charts/rstudio-connect/templates/NOTES.txt
ianpittwood, line 63: Should this check be extended to `backend.kubernetes`?

charts/rstudio-connect/templates/_helpers.tpl
```diff
@@ -155,3 +178,124 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
+    {{- if not (hasKey $jobBase.spec.template "spec") }}
+      {{- $_ := set $jobBase.spec.template "spec" (dict) }}
+    {{- end }}
+    {{- $podSpec := $jobBase.spec.template.spec }}
+    {{- /* add init container if not already present */}}
+    {{- $existingInits := default list $podSpec.initContainers }}
+    {{- $hasInit := false }}
+    {{- range $existingInits }}
+      {{- if eq .name "connect-content-init" }}
+        {{- $_ := set . "image" $image }}
```
ianpittwood: Should this be done conditionally as is done with the other elements? As it stands, this seems like it would clobber user values for the image being used.

charts/rstudio-connect/templates/rbac.yaml
```diff
@@ -1,14 +1,147 @@
...
+      - "create"
+  - apiGroups:
+      - ""
+    resources:
+      - "services"
+    verbs:
+      - "create"
+      - "get"
+      - "list"
+      - "watch"
```
ianpittwood, line 114: Missing delete here, not sure if it's intentional for it to be able to create services but never delete them.

charts/rstudio-connect/tests/kubernetes_test.yaml
ianpittwood, line 1: Thank you for writing some tests!

charts/rstudio-connect/templates/rbac.yaml
ianpittwood, line 1: I'm not seeing any tests validating the RBAC rules match expectations, might be worth adding.

charts/rstudio-connect/tests/kubernetes_test.yaml
```diff
@@ -0,0 +1,908 @@
...
+        enabled: true
+      backends:
+        kubernetes:
+          enabled: true
+    asserts:
+      - failedTemplate:
+          errorPattern: "launcher.enabled.*and.*backends.kubernetes.enabled.*cannot both be true"
+
+  # Test backends.kubernetes.enabled creates configmap entries for job base
+  - it: should create job.yaml in configmap when backends.kubernetes.enabled
```
ianpittwood, line 103: I'm not seeing a similar test for `service.yaml` being in the configmap

charts/rstudio-connect/values.yaml
ianpittwood: Should this be modified to include `backend.kubernetes`?

charts/rstudio-connect/templates/_helpers.tpl
ianpittwood, line 65: Expand comment to include `backend.kubernetes`

charts/rstudio-connect/values.yaml
```diff
@@ -299,6 +300,48 @@ chronicleAgent:
...
+      # -- The securityContext for the default initContainer
+      securityContext: {}
+    # -- defaultResourceJobBase is an optional Kubernetes Job definition used as the base
+    # when launching content jobs. The chart automatically adds the init container and
+    # runtime volume when backends.kubernetes.defaultInitContainer.enabled is true.
+    # Only set this if you need to customize the job (e.g., add sidecars, node selectors, tolerations).
+    # https://kubernetes.io/docs/concepts/workloads/controllers/job/
+    defaultResourceJobBase: {}
+    # -- defaultResourceServiceBase contains the Kubernetes Service definition which is used as an overlay "base" when creating a content job's
+    # Service in Kubernetes. Conceptually this is a similar to a Kustomize base. Connect then applies any required Service configuration
```
ianpittwood: nit: typo
```suggestion
    # Service in Kubernetes. Conceptually this is similar to a Kustomize base. Connect then applies any required Service configuration
```

charts/rstudio-connect/README.md.gotmpl
```diff
@@ -13,7 +13,7 @@
 ### 0.8.0
```
ianpittwood, line 17: Should we add an upgrade guidance section for 0.9.0?

charts/rstudio-connect/templates/deployment.yaml
```diff
@@ -219,6 +229,8 @@ spec:
           - name: overrides
             mountPath: "/mnt/job-json-overrides/"
           {{- end }}
+          {{- end }}
+          {{- if or .Values.launcher.enabled .Values.backends.kubernetes.enabled }}
```
ianpittwood: nit: we reference this conditional a lot, it could be a named helper like `oheEnabled`

ianpittwood (changes requested):
Some of these seem like potential bugs, please check them over and I'll give it another pass.

charts/rstudio-connect/values.yaml
lucasrod16: Good catch. Added validation to fail the deployment if these values conflict and tests

charts/rstudio-connect/templates/NOTES.txt
lucasrod16: Good catch. Added that check for `backend.kubernetes` and a test

charts/rstudio-connect/templates/_helpers.tpl
lucasrod16: Good catch. The `image` field now uses the same `hasKey` guard as the other field, so if a user explicitly sets `image` on their `connect-content-init` in `defaultResourceJobBase` it's preserved. Otherwise the chart sets it from `defaultInitContainer.repository/tag`

charts/rstudio-connect/templates/rbac.yaml
lucasrod16: This is intentional. The direct runner uses cascading deletion, so when Connect deletes the job Kubernetes automatically deletes the Service also via `ownerReference`.

lucasrod16: So Connect doesn't need that permission since it doesn't make that API call directly. Kubernetes handles it for us

charts/rstudio-connect/templates/rbac.yaml
lucasrod16: Good catch. Added a test to validate RBAC rules

charts/rstudio-connect/tests/kubernetes_test.yaml
lucasrod16: Good catch. Added a test for the `service.yaml` configmap

charts/rstudio-connect/values.yaml
lucasrod16: Good catch. Included `backends.kubernetes`

charts/rstudio-connect/templates/_helpers.tpl
lucasrod16: Done

charts/rstudio-connect/README.md.gotmpl
lucasrod16: Good call. Added upgrade guidance for 0.9.0 and linked to the upgrade from launcher to direct runner docs

charts/rstudio-connect/templates/deployment.yaml
lucasrod16: refactored to use an `oheEnabled` helper

lucasrod16:
@ianpittwood thanks for the review! It's ready for another look

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
ianpittwood, line 1: We should put some sort of callout in this file for the minimum Connect version that supports this feature.

charts/rstudio-connect/templates/_helpers.tpl
```diff
@@ -155,3 +187,126 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
+  then adds the init container, content container, and shared volume
+  when defaultInitContainer.enabled is true.
+*/}}
+{{- define "rstudio-connect.jobBase" -}}
+  {{- $jobBase := deepCopy (default (dict) .Values.backends.kubernetes.defaultResourceJobBase) }}
+  {{- /* ensure top-level fields */}}
+  {{- if not (hasKey $jobBase "apiVersion") }}
+    {{- $_ := set $jobBase "apiVersion" "batch/v1" }}
+  {{- end }}
+  {{- if not (hasKey $jobBase "kind") }}
```
ianpittwood, line 203: We should consider validating that `defaultResourceJobBase` and `defaultResourceServiceBase` have the correct `kind` associated with them. I can foresee some confusing errors propagating if a user makes a copy/paste error.

ianpittwood (approved):
Some smaller suggestions, but otherwise I think this is in a good state.

examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd
lucasrod16: Done

ianpittwood (approved):

PR #794, opened 2026-03-02
Support for adding annotations to deployments in Connect

### LINKED ISSUE #793: Support for adding annotations to deployment spec to Connect
## Description

Users would like the ability to add annotations to the deployment spec. Right now it's possible to add annotations to the pod spec and to the service spec, but not the deployment spec.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    reloader.stakater.com/auto: "true"
```

This request arose because I was talking to a user who is using a tool (https://github.com/stakater/Reloader) to automate redeployments when a secret changes. But that tool depends on an annotation being set on a deployment.

## Approach

- Add new top level item to values.yaml:

```yaml
deployment:
    annotations: {}
```

- Update the deployment template (https://github.com/rstudio/helm/blob/main/charts/rstudio-connect/templates/deployment.yaml) to add annotations set in values.yaml.

## Other helm charts

If we would like to support this ability, we should add it to Workbench and Package Manager as well. I created an issue on Connect as a starting point to get the approach correct.

SamEdwardes:
Closes #793 

---

Hey Posit Connect team, could you please review this PR. Details on the use case or in the issue mentioned above.

I deployed Connect using this chart and verified that the annotations correctly get set on the deployment:

```yaml
# values.yaml
fullnameOverride: posit-connect

deployment:
  annotations:
    deployment-annotate: Nice its working
```

```bash
kubectl get deployment posit-connect -o json | jq '.metadata.annotations'
{
  "deployment-annotate": "Nice its working",
  "deployment.kubernetes.io/revision": "18",
  "meta.helm.sh/release-name": "posit-connect",
  "meta.helm.sh/release-namespace": "samedwardes-posit-team"
}
```

### DIFF
--- a/.github/workflows/chart-test.yaml
+++ b/.github/workflows/chart-test.yaml
@@ -240,6 +240,9 @@ jobs:
       - name: Set up Helm
         uses: azure/setup-helm@v4
 
+      - name: Add rstudio helm repo
+        run: helm repo add rstudio https://helm.rstudio.com
+
       - uses: extractions/setup-just@v2
 
       - name: Run executable verification for default interpreters
--- a/Justfile
+++ b/Justfile
@@ -88,12 +88,13 @@ test chart='all':
 
 test-connect-interpreter-versions:
   #!/usr/bin/env bash
-  set -xe
+  set -euo pipefail
   cd ./charts/rstudio-connect && helm dependency build && cd -
 
   # find the default image
   image=$(
     helm template ./charts/rstudio-connect \
+    --set launcher.enabled=false \
     --show-only templates/deployment.yaml | \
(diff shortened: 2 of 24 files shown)


charts/rstudio-connect/NEWS.md
SamEdwardes: My editor auto-formatted on save, which caused the unintentional changes to older versions. These are purely formatting fixes, the only functional change is the new line added for 0.8.31.

SamEdwardes:
Thank you @lucasrod16. I added some unit tests. They are passing. Please let me know if you have any other comments, or if the PR is ready to merge.

Issue #295, opened 2022-11-07
Issue with launcher templates


pmaterer:
I'm trying to have my session pods use a specific service account (with an IRSA role).

I believe the way to do this is set it in `launcher.templateValues.pod.serviceAccoutName` (and have `launcher.useTemplates = true`)?

Here is are my values:

```
USER-SUPPLIED VALUES:
config:
  pam:
    common-session: |
      session [default=1] pam_permit.so
      session requisite   pam_deny.so
      session required    pam_permit.so
      session optional    pam_umask.so
      session required    pam_unix.so
    rstudio-session: |
      auth       sufficient pam_rootok.so
      auth       optional   pam_group.so
      session    required   pam_mkhomedir.so skel=/etc/skel/ umask=0022
      session    required   pam_env.so readenv=1 envfile=/etc/default/locale
      session    required   pam_limits.so
      session    optional   pam_sss.so
      @include common-auth
      @include common-account
      @include common-session
  profiles:
    launcher.kubernetes.profiles.conf:
      '*':
        allow-unknown-images: 0
        container-images: [redacted] # based on rstudio/r-session-complete:bionic-2022.07.1-554.pro3
        default-container-image: [redacted] # based on rstudio/r-session-complete:bionic-2022.07.1-554.pro3
        default-cpus: 1
        default-mem-mb: 8096
        max-cpus: 32
        max-mem-mb: 131072
    profiles:
      '*':
        session-limit: 4
  server:
    jupyter.conf:
      default-session-cluster: Kubernetes
      jupyter-exe: /usr/local/bin/jupyter
      labs-enabled: 1
      notebooks-enabled: 1
      session-cull-connected: 0
      session-cull-minutes: 540
      session-shutdown-minutes: 0
    launcher.conf:
      cluster:
        name: Kubernetes
        type: Kubernetes
    logging.conf:
      '*':
        log-level: info
        logger-type: stderr
    rserver.conf:
      admin-enabled: 1
      admin-superuser-group: rstudio-admins
      auth-pam-sessions-enabled: 1
      auth-pam-sessions-profile: rstudio-session
      auth-saml: 1
      auth-saml-metadata-url: [redacted]
      auth-saml-sp-attribute-username: NameID
      auth-saml-sp-base-uri: [redacted]
      launcher-address: 127.0.0.1
      launcher-default-cluster: Kubernetes
      launcher-port: 5559
      launcher-sessions-create-container-user: 1
      launcher-sessions-enabled: 1
[...]

PR #795, opened 2026-03-06
fix: prevent YAML line-wrap in job.tpl name/service_ports annotations

astayleraz:
## Summary

- Add `| indent 8 | trimPrefix (repeat 8 " ")` to the `name` and `service_ports` pod annotations in `job.tpl`
- Fixes invalid YAML when session names exceed Go's 80-character YAML marshal line-wrap threshold
- Matches the existing pattern used by `stdin` and `user_metadata` on adjacent lines

## Problem

Go's `yaml.Marshal` (used by `toYaml`) wraps strings at 80 characters. The `name` annotation renders the full session name including IDs, username, and project name. When this exceeds ~80 chars (e.g., with longer project names), the wrapped continuation line has indentation that doesn't match the surrounding YAML context. The K8s API rejects this with `400 Bad Request: could not find expected ':'`.

The `service_ports` annotation has the same latent vulnerability — its JSON payload can exceed 80 chars with multiple port mappings.

Same class of bug as #295.

## Test plan

- [ ] Deploy with a project name long enough to push the session name past 80 chars (e.g., "chronicle-reports-exploration")
- [ ] Verify session launches successfully
- [ ] Verify short project names still work (no regression)

Internal ref: Reported via #workbench Slack

### DIFF
--- a/charts/rstudio-workbench/README.md
+++ b/charts/rstudio-workbench/README.md
@@ -1,6 +1,6 @@
 # Posit Workbench
 
-![Version: 0.10.7](https://img.shields.io/badge/Version-0.10.7-informational?style=flat-square) ![AppVersion: 2026.01.1](https://img.shields.io/badge/AppVersion-2026.01.1-informational?style=flat-square)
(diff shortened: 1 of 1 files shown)


GCRev (approved):
LGTM

PR #796, opened 2026-03-06
Fix ip address handling.

ssinnott:
So there's a bug over here around ipv4 vs ipv6 handling.

`KUBERNETES_SERVICE_HOST` comes in as something like `2001:db8:1234:5678::1` the parsing below doesn't work because when it's concatenated with `KUBERNETES_SERVICE_PORT` it ends up with something like,

`https://2001:db8:1234:5678::1:443`

This checks if `:` is in the `KUBERNETES_SERVICE_HOST` if it is it assume ipv6 and properly places `[]`. So you end up with something like this,

`https://[2001:db8:1234:5678::1]:443`


ssinnott:
Verified the fix on our EKS ipv6 only environment.

PR #815, opened 2026-03-20
update helm chart overview

mleary:
Migrating information from older support articles into product documentation. Since this [article](https://support.posit.co/hc/en-us/articles/33089254334871-Posit-Helm-Cheatsheet) gave a broad overview that crosses product, it seemed reasonable to add it to the helm overview and point users to the product specific overviews with more details.

## Testing

- Tested this locally with `quarto preview`. I ran just docs per contributing guidelines but given this didnt touch charts/* I don't think it was needed.

- One action is failing. I believe this is due to the repo being a fork and not having access to secrets, can adjust as needed.

### DIFF
--- a/index.qmd
+++ b/index.qmd
@@ -23,53 +23,19 @@ title: Posit Helm Charts
 
 ## Installing Charts
 
-Example instructions to install a chart for each product are below. For full installation details, see the overview page for each product.
-
-::: {.panel-tabset}
-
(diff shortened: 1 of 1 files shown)


mleary:
@rstudio/infraops Bumping this up and hoping this is an easy PR!  We want to close out moving over support articles. thanks!

PR #818, opened 2026-03-26
pwb: Add top-level components key with Positron

bschwedler:
## Summary

Default the chart to using `workbench-session` and `workbench-session-init` containers, and add support for updating the Positron IDE version independently of a Workbench release.

- Add top-level `components` key with `enabled` toggle (default `true`) for session component init containers
- Configure `rserver.conf` with `launcher-sessions-auto-update` and `launcher-sessions-init-container-image-{name,tag}` when both launcher and components are enabled
- Add `components.positron` for Positron init container; set `version` to enable, `image.tag` to override for custom builds
- Positron init container attaches to server pod and configures `rserver.conf` for session pod injection
- When launcher is disabled, components are silently ignored
- To revert to the previous behavior, set `session.image.repository: rstudio/r-session-complete` and `components.enabled: false`
- Bump chart version to `0.11.0` (breaking change)

## Breaking Changes

Default to use the session init container.


ssinnott (dismissed):

skyeturriff (dismissed):
this looks awesome! just one minor comment!

zachhannum:
For the Workbench pod, instead of relying on Workbench's Positron scanning, the Helm chart should explicitly set the `exe` option in `positron.conf` automatically to the path the Helm chart expects to be created based on the init container being used. That way we guarantee that the version being put on session pods matches the version being used by the Workbench pod (largely this matters for hosting the correct version of the Positron docs).

charts/rstudio-workbench/templates/configmap-general.yaml
```diff
@@ -49,7 +49,26 @@
...
+    {{- $_ := set $defaultRServerConfigValues "launcher-positron-init-container-image-name" .Values.components.positron.image.repository }}
+    {{- $_ := set $defaultRServerConfigValues "launcher-positron-init-container-image-tag" $positronTag }}
+  {{- end }}
+{{- end }}
 {{- $defaultRServerConfig := dict "rserver.conf" ($defaultRServerConfigValues) }}
+{{- /* Point positron.conf exe at the version delivered by the positron init container
+    so the Workbench pod uses the same Positron version (and docs) as session pods. */}}
+{{- $defaultPositronConfValues := dict }}
+{{- if and .Values.launcher.enabled .Values.components.enabled .Values.components.positron.version }}
+  {{- $_ := set $defaultPositronConfValues "exe" (printf "/usr/lib/rstudio-server/bin/positron-server/%s/bin/positron-server/bundled/bin/positron-server" .Values.components.positron.version) }}
```
bschwedler: This `exe` path is deeply nested because the mount has no `subPath`.

Docs investigation (rstudio-pro `src/cpp/server/positron/`):
- `positron.conf` has an undocumented `docs-path` option
- If unset, Positron auto-discovers docs at `<exe_parent_parent>/docs`
- With current mount, auto-discovery looks at `<version>/bin/positron-server/bundled/docs` — wrong; docs land at `<version>/docs/positron/`
- Workbench serves them at `/docs/positron/` via nginx

Suggested alternative in `_helpers.tpl`: two `subPath` mounts of `positron-components` to line up with auto-discovery:

```yaml
- name: positron-components
  mountPath: /usr/lib/rstudio-server/bin/positron-server/{{ .Values.components.positron.version }}
  subPath: bin/positron-server/bundled
- name: positron-components
  mountPath: /usr/lib/rstudio-server/bin/positron-server/{{ .Values.components.positron.version }}/docs
  subPath: docs/positron/bundled
```

Result:
- `exe = /usr/lib/rstudio-server/bin/positron-server/<version>/bin/positron-server` (clean, mirrors the default `bundled/bin/positron-server` layout)
- Docs auto-discovered at `<version>/docs` — no need to set `docs-path`

`components.positron` isn't released yet, so this isn't breaking.

skyeturriff: Thanks for catching this. I like the `subPath` solution! IIUC, the alternative would be to make the positron init container "smarter", either by restructuring the path layout in the Dockerfile, or updating the Go binary to copy specific subdirs to specific target.

One note: the init container Dockerfile unzips the docs into `/opt/positron/docs/positron/` (with no `bundled/` dir). So I think we want
```
- name: positron-components
  mountPath: /usr/lib/rstudio-server/bin/positron-server/{{ .Values.components.positron.version }}
  subPath: bin/positron-server/bundled
- name: positron-components
  mountPath: /usr/lib/rstudio-server/bin/positron-server/{{ .Values.components.positron.version }}/docs
  subPath: docs/positron
```

charts/rstudio-workbench/tests/configmap_test.yaml
```diff
@@ -0,0 +1,238 @@
...
+    template: configmap-general.yaml
+    documentIndex: 0
+    set:
+      components:
+        positron:
+          version: "2026.03.0"
+    asserts:
+      - matchRegex:
+          path: data["positron.conf"]
+          pattern: "exe=/usr/lib/rstudio-server/bin/positron-server/2026\\.03\\.0/bin/positron-server/bundled/bin/positron-server"
```
skyeturriff: You were probably waiting to hear back on the `subPath` stuff but just in case - with the `subPath` stuff I think the pattern should be updated to 
```
pattern: "(?m)^exe=/usr/lib/rstudio-server/bin/positron-server/2026\\.03\\.0/bin/positron-server$"
```

same for the "should set positron.conf exe even when launcher is disabled" test below

charts/rstudio-workbench/tests/deployment_test.yaml
```diff
@@ -728,3 +728,108 @@ tests:
...
+        positron:
+          version: "2026.03.0"
+          image:
+            repository: "my-registry/positron-init"
+    asserts:
+      - equal:
+          path: 'spec.template.spec.initContainers[?(@.name=="positron-init")].image'
+          value: "my-registry/positron-init:2026.03.0"
+
+  - it: should add positron volume mount to server container
```
skyeturriff: just a note to update this for the `subPath` solution. could we do something like

```
- it: should add positron binary mount to server container
  template: deployment.yaml
  set:
    components:
      positron:
        version: "2026.03.0"
  asserts:
    - contains:
        path: 'spec.template.spec.containers[0].volumeMounts'
        content:
          name: "positron-components"
          mountPath: "/usr/lib/rstudio-server/bin/positron-server/2026.03.0"
          subPath: "bin/positron-server/bundled"
        any: true

- it: should add positron docs mount to server container
  template: deployment.yaml
  set:
    components:
      positron:
        version: "2026.03.0"
  asserts:
    - contains:
        path: 'spec.template.spec.containers[0].volumeMounts'
        content:
          name: "positron-components"
          mountPath: "/usr/lib/rstudio-server/bin/positron-server/2026.03.0/docs"
          subPath: "docs/positron"
        any: true
```

charts/rstudio-workbench/templates/configmap-general.yaml
```diff
@@ -49,7 +49,26 @@
...
+    {{- $_ := set $defaultRServerConfigValues "launcher-positron-init-container-image-name" .Values.components.positron.image.repository }}
+    {{- $_ := set $defaultRServerConfigValues "launcher-positron-init-container-image-tag" $positronTag }}
+  {{- end }}
+{{- end }}
 {{- $defaultRServerConfig := dict "rserver.conf" ($defaultRServerConfigValues) }}
+{{- /* Point positron.conf exe at the version delivered by the positron init container
+    so the Workbench pod uses the same Positron version (and docs) as session pods. */}}
+{{- $defaultPositronConfValues := dict }}
+{{- if and .Values.components.enabled .Values.components.positron.version }}
+  {{- $_ := set $defaultPositronConfValues "exe" (printf "/usr/lib/rstudio-server/bin/positron-server/%s/bin/positron-server/bundled/bin/positron-server" .Values.components.positron.version) }}
```
jforest: The constructed `exe` path bakes in four segments of the `workbench-positron-init` container's internal layout: `{version}/bin/positron-server/bundled/bin/positron-server`. If the init container ever restructures where it drops the binary (e.g., removes the `bundled/` layer, adds arch-specific subdirs), this silently breaks Positron on the Workbench pod.

A couple of options:

- Have the init container emit a known marker file (or accept an env var like `PWB_POSITRON_EXE_PATH`) so the chart doesn't need to know the internal layout.
- At minimum, drop a comment here pointing at the init container's source so a future maintainer editing *there* gets a search-hit *here*.

charts/rstudio-workbench/templates/configmap-general.yaml
```diff
@@ -49,7 +49,26 @@
...
+    {{- $_ := set $defaultRServerConfigValues "launcher-positron-init-container-enabled" 1 }}
+    {{- $_ := set $defaultRServerConfigValues "launcher-positron-init-container-image-name" .Values.components.positron.image.repository }}
+    {{- $_ := set $defaultRServerConfigValues "launcher-positron-init-container-image-tag" $positronTag }}
+  {{- end }}
+{{- end }}
 {{- $defaultRServerConfig := dict "rserver.conf" ($defaultRServerConfigValues) }}
+{{- /* Point positron.conf exe at the version delivered by the positron init container
+    so the Workbench pod uses the same Positron version (and docs) as session pods. */}}
+{{- $defaultPositronConfValues := dict }}
+{{- if and .Values.components.enabled .Values.components.positron.version }}
```
jforest: The guard `{{- if and .Values.components.enabled .Values.components.positron.version }}` appears in four places: here, `_helpers.tpl:191` (volume mount), and `deployment.yaml` (twice — the parent `initContainers` gate and the positron-init container block).

Consider extracting a named template:

```gotpl
{{- define "rstudio-workbench.positronInitEnabled" -}}
{{- and .Values.components.enabled .Values.components.positron.version -}}
{{- end }}
```

Each call site becomes `{{- if (include "rstudio-workbench.positronInitEnabled" .) }}`. Small change, atomic edits if the gate ever evolves (e.g., a third condition added).

charts/rstudio-workbench/templates/_helpers.tpl
```diff
@@ -188,6 +188,10 @@ containers:
       mountPath: "/var/lib/rstudio-launcher/Kubernetes/service.tpl"
       subPath: "service.tpl"
     {{- end }}
+    {{- if and .Values.components.enabled .Values.components.positron.version }}
+    - name: positron-components
+      mountPath: {{ printf "/usr/lib/rstudio-server/bin/positron-server/%s" .Values.components.positron.version | quote }}
```
jforest: Worth a one-line comment stating the implicit contract: **the `workbench-positron-init` container must deliver binaries at the volume root**, because the main container mounts the same shared `emptyDir` at `/usr/lib/rstudio-server/bin/positron-server/{version}` while the init container mounts it at `/mnt/init` (see `deployment.yaml`).

A future init-container refactor that writes under a subdirectory would break this silently — the chart itself still renders fine. One-line comment here avoids the trap.

charts/rstudio-workbench/README.md.gotmpl
```diff
@@ -118,6 +118,39 @@ Alternatively, database passwords may be set during `helm install` with the foll
 
 `--set config.secret.'database\.conf'.password="<YOUR_PASSWORD_HERE>"`
 
+## Session images
```
jforest, line 121: The new "Session images" section is great for new users but doesn't explicitly address existing users upgrading from 0.10.x. A short migration block would save a round trip to `NEWS.md` + the Posit release notes:

- **Upgrading from 0.10.x with chart defaults?** No changes needed.
- **Pinning `session.image.repository: rstudio/r-session-complete` explicitly today?** Add `components.enabled: false` to preserve previous behavior.
- **Want to adopt the new init-container pattern?** Confirm your custom session image does *not* include Workbench session components (only language runtimes, system tools, etc.) to avoid the init container duplicating work.

Even a link to Posit's public Workbench upgrade guide from this README would help — right now nothing connects a Helm operator to the platform-level upgrade docs.

charts/rstudio-workbench/README.md.gotmpl
```diff
@@ -118,6 +118,39 @@ Alternatively, database passwords may be set during `helm install` with the foll
...
+To update Positron independently of a Workbench release, set a version under
+`components.positron`:
+
+```yaml
+components:
+  positron:
+    version: "2026.03.0"
+```
+
+This attaches a separate init container that delivers the specified Positron version
```
jforest: Worth one more sentence here: bumping `components.positron.version` modifies the Workbench pod's init container spec, which triggers a Workbench pod rollout — not a zero-downtime session-only operation. Users coming from "Positron is updated client-side, this just changes what new session pods get" framing might not expect their Workbench admin UI to bounce when they bump Positron.

charts/rstudio-workbench/NEWS.md
```diff
@@ -1,6 +1,22 @@
 # Changelog
 
 
+## 0.11.0
+
+- **BREAKING**: Change default session image from `rstudio/r-session-complete` to `rstudio/workbench-session` and enable session component delivery via init containers by default.
+  Set `components.enabled: false` and `session.image.repository: rstudio/r-session-complete` to restore the previous behavior
+- Add top-level `components` key for configuring session component init containers
```
jforest, line 15: The chart now unconditionally writes `launcher-sessions-auto-update=1` to `rserver.conf` when `components.enabled=true` + `launcher.enabled=true`. Users who had this key set manually (common — we do in `rstudio/aws-main-services` today) will end up with a duplicate entry on upgrade, and users who relied on it being unset by default will be surprised by the behavior change.

Worth an explicit NEWS bullet, e.g.:

> - The chart now manages the `rserver.conf `launcher-sessions-auto-update`` key when `components.enabled: true` and `launcher.enabled: true`. Remove any manual `config.server.'rserver.conf'.launcher-sessions-auto-update` override to avoid duplicate configuration.

charts/rstudio-workbench/tests/configmap_test.yaml
```diff
@@ -0,0 +1,238 @@
...
+    documentIndex: 0
+    set:
+      components:
+        enabled: false
+        positron:
+          version: "2026.03.0"
+    asserts:
+      - notMatchRegex:
+          path: data["positron.conf"]
+          pattern: "^exe="
```
jforest, line 238: Consider adding one combined "legacy opt-out" regression test exercising the full rollback combo in one place. Existing tests cover individual "not present" assertions, but nothing exercises the realistic scenario of a user upgrading from 0.10.15 and setting `components.enabled: false` + `session.image.repository: rstudio/r-session-complete` to preserve prior behavior — under that combo, the chart should emit *zero* new-in-0.11.0 constructs.

```yaml
  - it: legacy opt-out produces no new-in-0.11.0 constructs
    template: configmap-general.yaml
    documentIndex: 0
    set:
      components:
        enabled: false
      session:
        image:
          repository: "rstudio/r-session-complete"
    asserts:
      - notMatchRegex:
          path: data["rserver.conf"]
          pattern: "launcher-sessions-(auto-update|init-container)"
      - notMatchRegex:
          path: data["rserver.conf"]
          pattern: "launcher-positron-init-container"
      - notMatchRegex:
          path: data["positron.conf"]
          pattern: "^exe="
```

And a companion in `deployment_test.yaml` asserting no `positron-init` container and no `positron-components` volume. Protects against a future refactor that adds a new construct gated on only one of the two toggles.

jforest (commented):
Claude and I reviewed this and came up with a few comments.

charts/rstudio-workbench/tests/configmap_test.yaml
bschwedler: I added tests, but called them something else.

skyeturriff (approved):
lgtm!

ssinnott (approved):
Nice!

PR #830, opened 2026-04-07
Add product-release workflow dispatch

bschwedler:
This workflow automates updating helm charts for bumping a product
version.

Adds a workflow_dispatch workflow that product image repos can
trigger after successful builds to update Helm chart appVersions.

The workflow:
- Maps product names to chart directories and image refs
- Bumps appVersion and chart patch version in Chart.yaml
- Updates artifacthub.io/images annotation
- Adds a NEWS.md changelog entry
- Regenerates docs via `just docs`
- Creates a PR targeting main via peter-evans/create-pull-request

### DIFF
--- a/.github/workflows/product-release.yml
+++ b/.github/workflows/product-release.yml
@@ -1,0 +1,175 @@
+name: Product Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      product:
+        description: "Product name"
+        required: true
+        type: choice
+        options:
+          - connect
+          - workbench
+          - package-manager
+      version:
+        description: "Product version to set as appVersion (e.g. 2026.03.0)"
+        required: true
+        type: string
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  update:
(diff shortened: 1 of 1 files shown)


.github/workflows/product-release.yml
```diff
@@ -0,0 +1,163 @@
...
+    inputs:
+      product:
+        description: "Product name"
+        required: true
+        type: choice
+        options:
+          - connect
+          - workbench
+          - package-manager
+      app-version:
```
jspiewak: `version` or `product-version`?
`app-` feels off.

jspiewak (approved):

.github/workflows/product-release.yml
bschwedler: I think changing this to `version` makes sense.

The [appVersion](https://helm.sh/docs/topics/charts/#the-appversion-field) is really internal to the chart definition itself, plus we are modifying it to strip off the build metadata anyway.

jforest (approved):
🤘 Thanks for implementing this

PR #842, opened 2026-04-15
feat(workbench): add limitStdinAnnotation to suppress sensitive stdin pod annotation

michaelmayer2:
## Summary

The session pod template in `job.tpl` writes the full stdin payload into a Kubernetes pod annotation named `stdin`. This payload contains session environment variables — which in some deployments include sensitive values such as **private TLS certificates**.

Pod annotations are visible in:
- `kubectl describe pod`
- Kubernetes audit logs
- Operator reconcile logs
- Cluster monitoring tools

The annotation is **purely informational** and has no effect on session functionality. Stdin delivery to the container is controlled separately via the container spec's `stdin: true` / `stdinOnce: true` fields, which are unaffected by this change.

## Changes

- `charts/rstudio-workbench/files/job.tpl` — wrap the `stdin` annotation in a conditional block guarded by `$templateData.pod.limitStdinAnnotation`
- `charts/rstudio-workbench/values.yaml` — add `launcher.templateValues.pod.limitStdinAnnotation: false` (default preserves existing behaviour)

## Usage

Customers can suppress the annotation by setting in their `values.yaml`:

```yaml
launcher:
  templateValues:
    pod:
      limitStdinAnnotation: true
```

## Test plan

- [ ] Deploy Workbench with default values — confirm `stdin` annotation is still present on session pods (backwards compatible)
- [ ] Deploy with `limitStdinAnnotation: true` — confirm `stdin` annotation is absent from session pods
- [ ] Confirm R/Python sessions launch and function normally with the annotation suppressed

## Background

Reported by a customer running Workbench in a regulated environment (Bfarm/FDZ project) where annotation contents are surfaced in their monitoring tooling.


GCRev (approved):

charts/rstudio-workbench/values.yaml
```diff
@@ -157,6 +157,12 @@ launcher:
       affinity: {}
       nodeSelector: {}
       hostAliases: []
+      # -- Set to true to suppress the `stdin` pod annotation on session pods.
+      # The annotation mirrors the full stdin payload (session environment variables)
+      # into Kubernetes pod metadata, making it visible in kubectl describe, audit
+      # logs, and monitoring tools. This can expose sensitive values such as TLS
+      # certificates. The annotation is purely informational and safe to disable.
+      limitStdinAnnotation: false
```
GCRev: maybe 
```suggestion
      suppressStdinAnnotation: false
```

michaelmayer2:
@GCRev - great idea with respects to the naming - much better than what I had suggested.

PR #852, opened 2026-04-22
Auto-apply connect.posit.co/service-account label on chart-managed SA (#846)

### LINKED ISSUE #846: Auto-apply `connect.posit.co/service-account=true` label on chart-managed SA when `backends.kubernetes.enabled: true`
## Problem

When `backends.kubernetes.enabled: true`, content jobs won't launch unless the service account they run as has the `connect.posit.co/service-account=true` label. The [upgrade guide](https://docs.posit.co/helm/examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.html#important-notes) tells users to apply this label manually with `kubectl label sa ...`.

If the chart creates the SA itself, users still have to remember to add the label via `rbac.serviceAccount.labels`. If they forget, content jobs fail with no obvious error at install time.

## Proposal

When both of these are true, the chart should add the label automatically on the SA it creates:

- `rbac.serviceAccount.create: true`
- `backends.kubernetes.enabled: true`

Any labels the user sets via `rbac.serviceAccount.labels` would still apply alongside the auto-added one.

## What wouldn't change

- `rbac.serviceAccount.create: false` — chart doesn't own the SA, so labeling stays manual.
- Content jobs pinned to a different SA via `defaultResourceJobBase.spec.template.spec.serviceAccountName` — chart can't know if it manages that SA, so labeling stays manual for those cases.

## Why

The chart already auto-wires several settings when `backends.kubernetes.enabled: true` (e.g. `DefaultResourceJobBase`, `ExecutionEnvironments.ConfigFilePath`, Role/RoleBinding). Adding the label fits the same pattern, and missing it is a common gotcha that's hard to debug.

## Docs

The upgrade guide at `examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd` should be updated alongside the chart change to note that the chart handles labeling in the default case, and that the manual `kubectl label sa ...` step only applies to BYO SAs or SAs pinned via `serviceAccountName` in the job base.

lucasrod16:
## Summary

- When `backends.kubernetes.enabled: true` and `rbac.serviceAccount.create: true`, the chart now automatically applies `connect.posit.co/service-account=true` to the service account it manages. This removes a common gotcha: the direct Kubernetes runner silently refuses to launch content jobs on service accounts without the label, and there's no clear signal at install time.
- User-provided labels under `rbac.serviceAccount.labels` are merged in alongside the auto-applied label; user values win on conflict, so it's still possible to override.
- Upgrade guide (`examples/connect/upgrade-launcher-to-kubernetes/launcher-to-kubernetes.qmd`) now scopes the manual `kubectl label sa` step to the cases the chart can't own: BYO service accounts (`rbac.serviceAccount.create: false`) or SAs pinned via `defaultResourceJobBase.spec.template.spec.serviceAccountName`.

Closes #846.

## Test plan

- [x] `just test rstudio-connect` (106 tests pass, including 4 new tests covering auto-apply, coexistence with user labels, user override on conflict, and the BYO-SA path)
- [x] `helm lint --strict -f ci/kubernetes-values.yaml` — clean
- [x] `helm template -f ci/kubernetes-values.yaml` shows the label on the rendered ServiceAccount
- [x] `helm template -f ci/kubernetes-values.yaml --set rbac.serviceAccount.create=false` renders no SA (as expected, no label to apply)
- [x] `just docs` regenerated `charts/rstudio-connect/README.md`
- [x] Chart version bumped `0.9.1` → `0.9.2` and `NEWS.md` updated

### DIFF
--- a/charts/rstudio-connect/templates/rbac.yaml
+++ b/charts/rstudio-connect/templates/rbac.yaml
@@ -49,6 +49,8 @@ subjects:
     namespace: {{ $namespace }}
 {{- end }}
 {{- if $serviceAccountCreate }}
+{{- /* Content jobs will not launch unless the SA has connect.posit.co/service-account=true, so the chart applies it automatically on the SA it manages. User-provided labels win on conflict. */}}
+{{- $mergedLabels := merge (deepCopy ($serviceAccountLabels | default dict)) (dict "connect.posit.co/service-account" "true") }}
 ---
 apiVersion: v1
 kind: ServiceAccount
@@ -59,10 +61,8 @@ metadata:
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-  {{- with $serviceAccountLabels }}
   labels:
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
+    {{- toYaml $mergedLabels | nindent 4 }}
 {{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/charts/rstudio-connect/Chart.yaml
+++ b/charts/rstudio-connect/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-connect
 description: Official Helm chart for Posit Connect
-version: 0.9.1
+version: 0.9.2
 apiVersion: v2
 appVersion: 2026.03.1
 icon: https://raw.githubusercontent.com/rstudio/helm/main/images/posit-icon-fullcolor.svg
--- a/charts/rstudio-connect/NEWS.md
+++ b/charts/rstudio-connect/NEWS.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.9.2
+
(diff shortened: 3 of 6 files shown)


dbkegley:
Good catch, thank you!

charts/rstudio-connect/templates/rbac.yaml
```diff
@@ -49,6 +49,8 @@ subjects:
     namespace: {{ $namespace }}
 {{- end }}
 {{- if $serviceAccountCreate }}
+{{- /* Content jobs will not launch unless the SA has connect.posit.co/service-account=true, so the chart applies it automatically on the SA it manages. User-provided labels win on conflict. */}}
+{{- $mergedLabels := merge (deepCopy ($serviceAccountLabels | default dict)) (dict "connect.posit.co/service-account" "true") }}
```
dbkegley, line 53: This is applying the label to Connect's service account though, right? Not the content service account. Is that what we want to do?

If the service account doesn't have the label then the jobs should be falling back to the default service account, not failing. Are you observing failures when this label is missing?

dbkegley (commented):
I hadn't considered managing the default content service account in the helm chart but that could be helpful for simplifying the upgrade. I'm not sure this is applying the labels to the right service account, but I'm not opposed to adding this feature to the chart

charts/rstudio-connect/templates/rbac.yaml
lucasrod16: You're right, this doesn't label the content service account. The issue and this PR was created by claude and shouldn't have been marked as ready for review. Claude misunderstood service account labeling and flagged it as a follow up when testing on reference architectures and I had it create an issue and work on it with the intention of reviewing it more closely once a draft PR was up. Thanks for the review and apologies for the noise! I'm going to close this PR and the issue

PR #875, opened 2026-05-29
Add `get` verb to `pods/attach` in Connect service account role for new kubernetes backend

dbkegley:
Bump `rstudio-connect` chart version to 0.20.3 and update NEWS.md

Update the rstudio-connect chart service account RBAC to allow websocket connections which are used by the kubernetes backend to execute jobs remotely on k8s. Connect 2026.04 correctly falls back to SPDY on websocket connection upgrade failures. The missing RBAC ensured that all connections were falling back to SPDY. 

Connect 2026.05.0 introduced a regression which prevented fallback to SPDY, meaning all steaming connections now fail (exec/attach). This PR allows the initial websocket connection to succeed. The next release of Connect will fix the SPDY fallback

### DIFF
--- a/charts/rstudio-connect/NEWS.md
+++ b/charts/rstudio-connect/NEWS.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.20.3
+
+- Add `get` on `pods/attach` and `pods/exec` to the direct Kubernetes runner Role in `templates/rbac.yaml` to support attach/exec authorization checks used by Kubernetes clients to support websocket connection upgrades.
+
 ## 0.20.2
 
 - Bump Connect version to 2026.05.0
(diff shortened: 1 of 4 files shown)


charts/rstudio-connect/NEWS.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.20.3
+
+- Add `get` on `pods/attach` to the direct Kubernetes runner Role in `templates/rbac.yaml` to support attach/exec authorization checks used by Kubernetes clients.
```
lucasrod16: Should we include a note in the NEWS about websocket support being the reason why we're adding this permission?

plascaray (approved):

charts/rstudio-connect/templates/rbac.yaml
lucasrod16, line 112: Will we also need `get` for `pods/exec` for killing jobs?

dbkegley: Good catch, thanks!

lucasrod16 (approved):
LGTM

PR #879, opened 2026-06-02
Rootless

ssinnott:
So idea here is to wire up a rootless version of workbench in the helm charts,

1) There's a simple high level flag (`pod. runAsRoot`) this defaults to true (for now). But you can set it to false to basically wire in a pretty separate mode for workbench. Where,
   a. `sssd` is disabled - you need to use SCIM instead.
   b. Workbench and pods that it launches will use the service account instead.
   c. There's some other toggling which is more minor.

2) If you're in the rootless mode you can also specify the service account user and uid. For this to really work though I think you would need to build your own version of the workbench image. The user + uid is baked into there with my other changes.

Closes https://github.com/rstudio/rstudio-pro/issues/10847


charts/rstudio-workbench/tests/deployment_test.yaml
```diff
@@ -852,3 +857,155 @@ tests:
...
+          value: 1500
+
+  - it: should let pod.securityContext override the non-root defaults
+    template: deployment.yaml
+    set:
+      pod.runAsRoot: false
+      pod:
+        securityContext:
+          runAsUser: 4242
+          fsGroup: 4242
```
GCRev: Recommend making this unique, even if that's unrealistic, just to be sure these values end up where they are supposed to.
```suggestion
          fsGroup: 4243
```

charts/rstudio-workbench/tests/deployment_test.yaml
```diff
@@ -852,3 +857,155 @@ tests:
...
+    asserts:
+      - equal:
+          path: spec.template.spec.securityContext.runAsUser
+          value: 4242
+      - equal:
+          path: spec.template.spec.securityContext.fsGroup
+          value: 4242
```
GCRev:
```suggestion
          value: 4243
```

charts/rstudio-workbench/tests/deployment_test.yaml
ssinnott: good idea!

GCRev (approved):
If there are any problems with this, then I didn't spot them.

charts/rstudio-workbench/NEWS.md
```diff
@@ -1,9 +1,13 @@
...
-
-## 0.20.1
-
-- Bump Workbench version to 2026.05.0
+## 0.21.1
+
+- Workbench pods can now run as a non-root user. Set `pod.runAsRoot: false` (new value, default `true`) to run unprivileged.
+- When running unprivelaged SCIM must be used for user management.
+- A new `config.sssd` block controls the bundled SSSD daemon (replaces the deprecated `config.userProvisioning`).
+- `config.userProvisioning` is deprecated; use `config.sssd.conf`. A warning is emitted when it is set.
```
adamhigerd: Consider rewriting these two lines into a single bullet point; they feel kind of redundant. I'm also confused by `config.sssd` vs `config.sssd.conf`.

charts/rstudio-workbench/NEWS.md
```diff
@@ -1,9 +1,13 @@
 # Changelog
 
-
-## 0.20.1
-
-- Bump Workbench version to 2026.05.0
+## 0.21.1
+
+- Workbench pods can now run as a non-root user. Set `pod.runAsRoot: false` (new value, default `true`) to run unprivileged.
+- When running unprivelaged SCIM must be used for user management.
```
adamhigerd:
```suggestion
- Use SCIM for user management when running unprivileged.
```

charts/rstudio-workbench/NEWS.md
```diff
@@ -1,9 +1,13 @@
...
+- A new `config.sssd` block controls the bundled SSSD daemon (replaces the deprecated `config.userProvisioning`).
+- `config.userProvisioning` is deprecated; use `config.sssd.conf`. A warning is emitted when it is set.
+- The Workbench and launcher prestart scripts (`prestart-workbench.bash`, `prestart-launcher.bash`) no longer perform root-only operations.
```
adamhigerd: How relevant is this to the end admin? I'm not familiar enough to know if this is important to call out or if it could be dropped.

charts/rstudio-workbench/NEWS.md
```diff
@@ -1,9 +1,13 @@
...
+- A writable `emptyDir` is now mounted at `/mnt/load-balancer` when load balancing is active (`replicas > 1` or `loadBalancer.forceEnabled: true`), so the prestart script can write the load-balancer config when the pod runs unprivileged (`pod.runAsRoot: false`).
```
adamhigerd:
```suggestion
- A writable `emptyDir` is now mounted at `/mnt/load-balancer` when load balancing is active (`replicas > 1` or `loadBalancer.forceEnabled: true`). The prestart script writes the load-balancer config in this directory when the pod runs unprivileged (`pod.runAsRoot: false`).
```

(Or perhaps "can write" instead of "writes"; I don't know the details.)

charts/rstudio-workbench/README.md
```diff
@@ -302,9 +304,9 @@ the `XDG_CONFIG_DIRS` environment variable.
   - It is mounted into the pod at `/scripts/`.
   - `prestart-workbench.bash` is used to start workbench.
   - `prestart-launcher.bash` is used to start launcher.
-- User Provisioning Configuration:
-  - These configuration files are used for configuring user provisioning (i.e., `sssd`).
-  - Located at:<br> `config.userProvisioning.<< name of file >>` Helm values
+- SSSD Configuration:
+  - These configuration files configure the bundled `sssd` daemon (legacy LDAP/AD provisioning), which is started only when `config.sssd.enabled=true`.
+  - Located at:<br> `config.sssd.conf.<< name of file >>` Helm values (the deprecated `config.userProvisioning` is honored as a fallback)
```
adamhigerd: This line seems _really_ awkward... I'm not sure what you mean.

charts/rstudio-workbench/README.md
```diff
@@ -355,9 +357,9 @@ Provisioning users in Workbench containers is challenging. Session images create
 consistent UIDs / GIDs). However, creating users in the Workbench containers is a responsibility that falls to the
 administrator.
 
-The most common way to provision users is via `sssd`.
-The [latest Workbench container](https://github.com/rstudio/rstudio-docker-products/tree/main/workbench#user-provisioning)
-has `sssd` included and running by default (see `userProvisioning` configuration files above).
+Posit Workbench's native user provisioning (SCIM / just-in-time) is the recommended approach and does not require SSSD.
+The legacy approach is `sssd`: the [latest Workbench container](https://github.com/rstudio/rstudio-docker-products/tree/main/workbench#user-provisioning)
```
adamhigerd:
```suggestion
`sssd` is the legacy approach: the [latest Workbench container](https://github.com/rstudio/rstudio-docker-products/tree/main/workbench#user-provisioning)
```

charts/rstudio-workbench/README.md
```diff
@@ -616,7 +618,7 @@ Use of [Sealed secrets](https://github.com/bitnami-labs/sealed-secrets) disables
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` | A map used verbatim as the pod's "affinity" definition |
-| args | list | `[]` | args is the pod container's run arguments. |
+| args | list | `[]` | args is the pod container's run arguments. When unset, the container's default arguments are used. |
```
adamhigerd: Consider changing the content of the "default" column. Not sure what to suggest though.

charts/rstudio-workbench/README.md.gotmpl
```diff
@@ -33,20 +33,22 @@ To function, this chart requires the following:
 * If using load balancing (by setting `replicas > 1`), you need similar storage defined for `sharedStorage` to
   store shared project configuration. However, you can also configure the product to store its shared data underneath `/home` by
   setting `config.server.rserver\.conf.server-shared-storage-path=/home/some-shared-dir`.
-* A method to join the deployed `rstudio-workbench` container to your auth domain. The default `posit/workbench` image has `sssd` installed and started by default.
-  You can include `sssd` configuration in `config.userProvisioning` like so:
+* A method to join the deployed `rstudio-workbench` container to your auth domain. The `posit/workbench` image ships `sssd` for legacy LDAP/Active Directory provisioning. SSSD must run as root, so the bundled daemon starts by default only on root deployments (`config.sssd.enabled: true`, `pod.runAsRoot: true`) and is automatically skipped when `pod.runAsRoot: false`. Modern provisioning (SCIM / native) does not require SSSD.
+  Provide its configuration in `config.sssd.conf` like so (set `config.sssd.enabled: false` to disable the daemon entirely):
```
adamhigerd: I wonder if this is too much information?

adamhigerd:
I'm going to recuse myself from this PR beyond basic documentation style. I don't really know much about this subject matter and I'm heading out of town tomorrow.

charts/rstudio-workbench/NEWS.md
```diff
@@ -1,9 +1,17 @@
...
+- In rootless mode (`pod.runAsRoot: false`), the chart-managed `secure-cookie-key` is now projected into `/mnt/secret-configmap/rstudio/` even when it is autogenerated. The rootless `launcher.conf` default points `secure-cookie-key-file` there so the launcher shares rserver's cookie key; previously the file was unmounted with the default key and the launcher read a missing file.
+- The launcher prestart script now creates its scratch dirs (`/var/lib/rstudio-launcher/Local`, `/Kubernetes`) setgid and group-writable, so the launcher (which runs as the non-root `serviceAccountUser`) can write to them in root-mode pods without a privileged ownership change.
+
+- Workbench pods can now run as a non-root user. Set `pod.runAsRoot: false` (new value, default `true`) to run unprivileged.
+- When running unprivelaged SCIM must be used for user management.
```
zachhannum: I still don't fully understand how SCIM works here but JIT doesn't. Fundamentally it seems like the requirement is that Workbench native user provisioning must be used, not necessarily the exact provisioning method?

charts/rstudio-workbench/values.yaml
```diff
@@ -404,6 +404,18 @@ podDisruptionBudget: {}
 revisionHistoryLimit: 3
 
 pod:
+  # -- Whether the pod's containers run as the root OS user. `true` (default) preserves historical
+  # behavior: pod runs as root (`runAsUser: 0`), SSSD starts, secrets mount 0600, PAM sessions on.
+  # `false` runs unprivileged: pod sets `runAsNonRoot: true` and `runAsUser`/`fsGroup` to
+  # `pod.serviceAccountUserId`, SSSD is skipped, secrets mount 0640, and non-root
+  # `rserver.conf`/`launcher.conf` defaults are applied.
+  runAsRoot: true
```
zachhannum, line 413: Is making this `true` by default a breaking change?

charts/rstudio-workbench/NEWS.md
ssinnott: Ah! Yeah I'm conflating the two here. I'll be clearer. This work with SCIM as well as JIT. You need to use openid/saml or some other auth provider - local auth basically doesn't work because that leans on pam.

charts/rstudio-workbench/values.yaml
ssinnott: By default we currently run as root - flipping this to false will enabled the rootless mode. Which is beaking.

zachhannum: 🤦 Duh. Ignore me.

PR #884, opened 2026-06-04
Add fork-safe install CI with replicas=0

bschwedler:
Fork PRs have no access to repo secrets, so `ct install` — which waits for pods that require a valid license to start — cannot run as-is against a fork PR.

- Add a `detect-fork` job that outputs `is-fork` using `github.event.pull_request.head.repo.full_name != github.repository`. The expression is assigned via `env:` before use in `run:` to prevent script injection.
- Gate `Create License File Secrets` on non-fork PRs.
- Split `ct install` into two pairs: the existing pair (license-backed, non-fork only) and a new fork pair using `--helm-extra-set-args "--set=replicas=0"`.
- With `replicas=0`, Kubernetes creates all non-pod resources (Services, RBAC, ConfigMaps, PVCs) and considers the Deployment immediately rolled out — no pod is ever scheduled that would require a license.
- Update the failure gate to include all four install step outcomes.

Internal PRs are completely unchanged. Fork PRs get Kubernetes resource validation without needing license secrets.

### DIFF
--- a/.github/workflows/chart-test.yaml
+++ b/.github/workflows/chart-test.yaml
@@ -87,7 +87,7 @@ jobs:
           version: "v3.19.2"
 
       - name: Install helm unittest plugin
-        run: helm plugin install https://github.com/helm-unittest/helm-unittest.git --version v1.0.0
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest.git --version v1.1.0
 
       - name: Run chart unit tests
         id: unittest
--- a/.github/workflows/chart-test.yaml
+++ b/.github/workflows/chart-test.yaml
@@ -137,8 +137,21 @@ jobs:
         if: steps.unittest.outcome == 'failure'
         run: exit 1
 
+  detect-fork:
+    runs-on: ubuntu-latest
+    permissions: {}
+    outputs:
+      is-fork: ${{ steps.check.outputs.is-fork }}
+    steps:
+      - name: Check fork status
+        id: check
+        env:
+          IS_FORK: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository }}
+        run: echo "is-fork=$IS_FORK" >> "$GITHUB_OUTPUT"
+
   install:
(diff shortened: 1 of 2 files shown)


matt-urbina (approved):

bschwedler:
@jforest What are your thoughts on this change? I can most definitely rebase.

.github/workflows/chart-test.yaml
```diff
@@ -201,16 +215,32 @@ jobs:
...
         run: ct install --target-branch main --all --chart-dirs charts --chart-dirs other-charts --excluded-charts rstudio-library,rstudio-library-test --namespace posit-test
         continue-on-error: true
 
+      # Fork PRs have no access to license secrets. replicas=0 causes Kubernetes to
+      # create all non-pod resources (Services, RBAC, ConfigMaps, PVCs) and consider
+      # the Deployment immediately rolled out — no pod is scheduled that needs the license.
+      - name: Run chart-testing (install changed, fork)
+        id: ct-install-fork
+        if: ${{ github.ref != 'refs/heads/main' && steps.list-changed.outputs.changed == 'true' && needs.detect-fork.outputs.is-fork == 'true' }}
+        run: ct install --target-branch main --chart-dirs charts --chart-dirs other-charts --excluded-charts rstudio-library,rstudio-library-test --namespace posit-test --helm-extra-set-args "--set=replicas=0"
```
jforest, line 235: Is there value to us having replicas for any of the tests?  IE: do we need replicas when we run our own PRs?

if not lets just default to that and skip all this

bschwedler: @jforest If we deploy no replicas, the install tests never actually start the container.

It comes down to what do we want to be testing as part of CI.

bschwedler: I think we need the actual install to test whether the container pulls correctly.

jforest: Fair, then it soundsl ike this is the way

jforest (approved):

PR #894, opened 2026-06-17
remove chronicle charts

markrtucker:
With the 2026.06 releases, Chronicle no longer runs as a separate pod. The chronicle process is managed by Workbench and/or Connect, and is controlled by product-specific configuration values. This PR removes all chronicle-specific charts and configuration.

This should not be merged until after the 2026.06 releases.

I'm not sure about the chart versioning practices and if these should be minor or major version bumps.


jspiewak:
Won't this break anyone who currently deploys Chronicle via this helm chart?
Shouldn't you deprecate the chart first, and keep it for a long time in order to give users (including CloudOps) time to migrate to your new way of deploying?

markrtucker:
> Won't this break anyone who currently deploys Chronicle via this helm chart? Shouldn't you deprecate the chart first, and keep it for a long time in order to give users (including CloudOps) time to migrate to your new way of deploying?

That's a fair point. I was initially thinking that users could just continue to use older versions of the charts, but that potentially blocks users from picking up new chart features.

@bschwedler or @ianpittwood - do we have any recommended approaches on how to mark charts as deprecated?

bschwedler:
@markrtucker We will need to release a version of the chart that is marked as `deprecated`.  The following need to be added to the `Chart.yaml`

The Chart itself: https://helm.sh/docs/topics/charts/#the-chartyaml-file
```yaml
deprecated: true
```

We will also need to update the README template & news, as well as updating references to the chart.

bschwedler:
> I'm not sure about the chart versioning practices and if these should be minor or major version bumps.

@markrtucker Since we are pre 1.0, this should be a minor version bump.

zachhannum:
Are we OK with pushing existing users of server Chronicle to move towards using Chronicle embedded in the products even though the feature is marked as Preview for the 2026.06 release?

markrtucker:
> Are we OK with pushing existing users of server Chronicle to move towards using Chronicle embedded in the products even though the feature is marked as Preview for the 2026.06 release?

I'll defer to @jvroberts for an official answer, but I think we are ok with this. We have very few "legacy" Chronicle customers, and a very small percentage of those run on k8s. We expect that post.it may be the biggest k8s based deployment of Chronicle.

More broadly, my plan is to close this PR and create a new PR that deprecates the chronicle charts.

markrtucker:
Closing in favor of https://github.com/rstudio/helm/pull/895

PR #896, opened 2026-06-19
move connect runtime mount location

karawoo:
This mounts the Connect runtime at `/opt/rstudio-connect-runtime` instead of `/opt/rstudio-connect` and sets `config.Kubernetes.ConnectRuntimeDir` to match, so the content mount and the path Connect looks at can't diverge.

The reason for this is that containerized content lives inside the built images at `/opt/rstudio-connect/mnt/app` but with the runtime mounted to `/opt/rstudio-connect` that path is obscured and containerized content fails to run.

Previous versions of the chart mounted the runtime components at subpaths under `/opt/rsudio-connect` (0.8.14 did this for the launcher), but I opted against that to avoid having to maintain separate mounts in the event the runtime directory structure changes.

Relocating the app content instead would also be a possibility but would break existing containerized applications that have already put their content at `mnt/app`.

### DIFF
--- a/charts/rstudio-connect/templates/_helpers.tpl
+++ b/charts/rstudio-connect/templates/_helpers.tpl
@@ -101,6 +101,13 @@ app.kubernetes.io/instance: {{ .Release.Name }}
     {{- if .Values.backends.kubernetes.defaultResourceServiceBase }}
       {{- $_ := set $kubernetesSettingsDict "DefaultResourceServiceBase" (default "/etc/rstudio-connect/service.yaml" (dig "Kubernetes" "DefaultResourceServiceBase" "" .Values.config)) }}
     {{- end }}
+    {{- /* When the default init container is enabled, default ConnectRuntimeDir to
+           the runtime mount path. This default MUST match the $runtimeDir default
+           in rstudio-connect.jobBase, which derives the content container's
+           mountPath from this same value so the config and mount cannot diverge. */}}
+    {{- if and .Values.backends.kubernetes.defaultInitContainer.enabled (not (dig "Kubernetes" "ConnectRuntimeDir" "" .Values.config)) }}
+      {{- $_ := set $kubernetesSettingsDict "ConnectRuntimeDir" "/opt/rstudio-connect-runtime" }}
+    {{- end }}
     {{- $kubernetesDict := dict "Kubernetes" ( $kubernetesSettingsDict ) }}
     {{- $defaultConfig = merge $defaultConfig $kubernetesDict }}
   {{- end }}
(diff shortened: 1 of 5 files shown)


dbkegley (approved):

charts/rstudio-connect/templates/_helpers.tpl
```diff
@@ -166,7 +173,13 @@ app.kubernetes.io/instance: {{ .Release.Name }}
       {{- $_ := set $initContainer "securityContext" .Values.backends.kubernetes.defaultInitContainer.securityContext }}
     {{- end }}
     {{- /* build the content container */}}
-    {{- $contentVolumeMount := dict "name" "rsc-volume" "mountPath" "/opt/rstudio-connect" }}
+    {{- /* The content container mounts the runtime volume at ConnectRuntimeDir,
+           where Connect expects to find the runtime components. Derive the
+           mountPath from the resolved config value so the two can never diverge.
+           The default here MUST match the ConnectRuntimeDir default injected in
+           rstudio-connect.config. */}}
+    {{- $runtimeDir := dig "Kubernetes" "ConnectRuntimeDir" "/opt/rstudio-connect-runtime" .Values.config }}
```
lucasrod16: Could we normalize empty/null values here? `dig` only applies the default when the key is missing, so if a user sets `config.Kubernetes.ConnectRuntimeDir: ""` (or null), this renders `mountPath: ""/null` in job.yaml. Idk how likely that is to happen, but guarding against that case might be worth it.


```suggestion
    {{- $runtimeDir := default "/opt/rstudio-connect-runtime" (dig "Kubernetes" "ConnectRuntimeDir" "" .Values.config) }}
```

charts/rstudio-connect/templates/_helpers.tpl
lucasrod16, line 138: Also consider unsetting empty `ConnectRuntimeDir` in rstudio-connect config so the rendered gcfg value stays aligned with the mount path. A unit test to ensure empty/null overrides fallback to the default path would be good to have as well.

Something like this added to this block:

``` 
{{- if not (dig "Kubernetes" "ConnectRuntimeDir" "" $configCopy) }}
   {{- $_ := unset (index $configCopy "Kubernetes") "ConnectRuntimeDir" }}
{{- end }}
```

charts/rstudio-connect/tests/kubernetes_test.yaml
```diff
@@ -838,7 +885,7 @@ tests:
                     - name: connect-content
                       volumeMounts:
                         - name: other-volume
-                          mountPath: /opt/rstudio-connect
+                          mountPath: /opt/rstudio-connect-runtime
```
lucasrod16, line 909: I think we want to keep test coverage to ensure `/opt/rstudio-connect` can't be mounted over since `/opt/rstudio-connect/mnt/app` is the runtime bind-mount path seen by content processes

lucasrod16 (approved):

PR #915, opened 2026-07-24
Pass launcher capabilities through to session container in job.tpl

ssinnott:
## Problem

Under `use-templating=1` the chart-owned session job template (`files/job.tpl`) fully owns the pod spec but only maps `runAsUser`/`runAsGroup`/`supplementalGroups`. It never reads `.Job.container.capabilitiesAdd` / `.Job.container.capabilitiesDrop`, so Linux capabilities rserver requests for the session container are silently dropped.

This breaks rootless sessions (`launcher-kubernetes-identity=server-user-capabilities`), which need `CHOWN`/`DAC_OVERRIDE`: `pwb-mkhomedir` starts with an empty capability set and fails with `missing required capability: cap_chown`. The launcher's native render (`use-templating=0`) already emits these caps; only full-templating deployments are affected. The default root identity sends no caps and is unaffected.

## Fix

Merge the launcher-sent capabilities into the container-level `securityContext`, alongside `launcher.templateValues.pod.containerSecurityContext` (launcher caps are concatenated onto any chart-configured `add`/`drop` lists), mirroring the launcher's built-in template. Merging rather than emitting a separate block avoids a duplicate `securityContext:` key when a deployment also sets `containerSecurityContext`.

When no caps are sent, rendering is unchanged.

## Testing

- Rendered `job.tpl` through a Go harness reproducing the launcher's templating (text/template + sprig + `toYaml`/`include`, with `rstudio-library.templates.data` defined as `configmap-general.yaml` ships it). Verified valid YAML and correct `securityContext` for: caps only; caps merged with an existing `containerSecurityContext`; no caps with/without `containerSecurityContext` (byte-identical to previous output).
- `just test rstudio-workbench`: 169/169 pass
- `helm lint --strict` with default values and every `ci/*.yaml`

### DIFF
--- a/charts/rstudio-workbench/Chart.yaml
+++ b/charts/rstudio-workbench/Chart.yaml
@@ -1,6 +1,6 @@
 name: rstudio-workbench
 description: Official Helm chart for Posit Workbench
-version: 0.21.4
+version: 0.21.5
 apiVersion: v2
 appVersion: 2026.07.1
 icon:
--- a/charts/rstudio-workbench/NEWS.md
+++ b/charts/rstudio-workbench/NEWS.md
@@ -3,7 +3,7 @@
 
 ## 0.21.5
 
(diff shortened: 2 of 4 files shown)


zachhannum (approved):
LGTM. Would it be possible / potentially better to natively handle this in the default launcher job template? e.g. in the launcher code itself.

ssinnott:
> LGTM. Would it be possible / potentially better to natively handle this in the default launcher job template? e.g. in the launcher code itself.

Great question! So I did make the launcher change as well that opens up this pathway for templating.

I ran into this in testing - but I think it's by design. The template here basically replaces the default template so if the caps aren't handled here they don't end up on the session.

GCRev (approved):

Issue #929, opened 2026-08-19
CI/workflow reliability fixes (Aug 2026)


bschwedler:
Tracking recent GitHub Actions workflow breakages and fixes, so recurring CI fragility is visible in one place instead of scattered across individual PRs.

## Fixed
- [x] Helm 4.2.4 changed manifest-splitting output, breaking the `rbac` CI check on every PR
  - https://github.com/rstudio/helm/pull/928
  - Not pinned to a Helm version, so a future Helm release could reintroduce this

## In progress
- [ ] Slack failure notifications silently broken since `slack-github-action` v3.0.3 bump — `webhook-type` must be an action input, not the `SLACK_WEBHOOK_TYPE` env var (affects chart-doc, chart-test, chart-releaser, chart-rebuild, publish workflows)

PR #931, opened 2026-08-19
Add zizmor pre-commit hook

bschwedler:
Tracked in https://github.com/rstudio/helm/issues/929

zizmor already runs in CI (chart-test.yaml), but findings only surface after a push. This adds it as a pre-commit hook so findings are caught locally, following the pattern from https://github.com/posit-dev/images-shared/pull/732.

Adding the hook required first getting the repo to a clean baseline:
- 5 high-confidence template-injection findings in issues.yml — `github.event.label.name` was interpolated directly into a shell `if` condition; a crafted label name could inject shell code. Moved into an env var.
- 10 checkouts missing `persist-credentials: false`. None of these jobs push using the checkout's ambient git credentials — writes go through `chart-releaser-action`, `create-pull-request`, or an explicit GitHub App token — so persisting the token in `.git/config` served no purpose.
- 8 jobs with no `permissions:` block, inheriting the default repo-wide `GITHUB_TOKEN`. Scoped each to what it actually uses.
- 2 low-confidence template-injection findings in chart-rebuild.yaml (PR number/URL echoed into a shell string) — same env var fix.

`.pre-commit-config.yaml` is new to this repo; it currently contains only the zizmor hook.

### DIFF
--- a/.github/workflows/chart-doc.yaml
+++ b/.github/workflows/chart-doc.yaml
@@ -8,9 +8,13 @@ jobs:
   document:
     runs-on: ubuntu-latest
     name: helm-docs
+    permissions:
+      contents: read
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          persist-credentials: false
 
       - name: Set up Helm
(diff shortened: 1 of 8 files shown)


jspiewak (approved):

jforest (approved):
🤘

